PR #19, opened 2015-12-24
Create table module

Narrev:
Adds several new functions to table!
Including:
table.contains()
table.random()
table.copy()


JetCrusherTorpedo:
I suggest renaming *table.check()* to *table.contains()*.

PR #20, opened 2015-12-24
os Perfected!

Narrev:
The os module is now 100% accurate and errors only when it is supposed to!

Yes, in some cases (when returning the long string of gsubs) the code does slightly favor optimization over readability. This is because a 5-16% increase in speed (less overhead) is more important than being able to read a function that should already be built-in.

If anyone wants to see exactly how these calculations work, I posted a readability view [here](https://github.com/Narrev/Unix-Epoch-Date-and-Time-Calculations/blob/master/Time%20Functions.lua).


Narrev:
With the new efficiency fix, the module will sometimes be inaccurate after the year 3476. It bugs me but doesn't actually matter for any practical use.

Narrev:
It is now accurate forever!

Narrev:
I am still working on the algorithm. Should have it perfected tomorrow

Narrev:
I guess the os module is as good as it is going to get unless I fix it for years after 3480. It actually still is almost always accurate after 3480. Before that though, it is completely precise. Message me for details if needed

Narrev:
The os module is always accurate when dealing with positive timestamps. I did not make it automatically error for negative timestamps, because sometimes it will work, but I wouldn't use negative inputs on a server.

Let me know if you think it would be better to make the date function error if given a negative input

PR #30, opened 2015-12-31
Sign function is now a single line

Narrev:
Shorter!


Quenty:
I'm not sure why you closed this, but I'd like to point out the `Sign` function is fine as multiple lines, as it is more readable this way.

PR #34, opened 2016-01-21
Added to "Make" and "Modify" functions

Narrev:
You can still use the "Make" function exactly as before, but now you can call it with extra tables which return an object for each subsequent table. The objects returned have the properties from Properties applied first, then anything in their table. Basically, you can create a list of default properties and make a bunch of objects at once that have the same properties.

### DIFF
--- a/Modules/qGUI/ScreenCover.lua
+++ b/Modules/qGUI/ScreenCover.lua
@@ -1,3 +1,7 @@
+-- ScreenCover.lua
+-- @author Quenty
+-- Last Modified November 17th, 2014
+
 local ReplicatedStorage = game:GetService("ReplicatedStorage")
 local RunService        = game:GetService("RunService")
 
@@ -10,20 +14,23 @@ local qMath             = LoadCustomLibrary("qMath");
 
 local lib               = {}
 
--- ScreenCover.lua
--- @author Quenty
--- Last Modified November 17th, 2014
+local Make              = qSystems.Make
+local Round             = qSystems.Round
+local Modify            = qSystems.Modify
+local roundUp           = qMath.roundUp
+local GetEnumName       = EnumInterpreter.GetEnumName
 
-local Make   = qSystems.Make
-local Round  = qSystems.Round
-local Modify = qSystems.Modify
+local max               = math.max
+local insert            = table.insert
 
+local UDim2             = UDim2.new
+local Color3            = Color3.new
 
-lib.STYLES = {}
-lib.TYPES = {
-	"TransitionOut";
-	"TransitionIn";
-}
+local STYLES = {}
+local TYPES = {"TransitionOut", "TransitionIn"}
+
+lib.STYLES = STYLES
+lib.TYPES  = TYPES
 
 local function MakeCover(Properties)
 	-- Generates a cover frame that is basically standard. :D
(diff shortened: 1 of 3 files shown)


Quenty:
I like the changes made here. I'm a little bit concerned about performance, but the actual act of instancing a new object in ROBLOX is so expensive that the added if-statement(s) are so minimal that this seems reasonable.

- Great job reducing linecount and removing code
- New API changes are good. 


However, there are a few (really small) things I'd like to change. I'll merge this pull request one the following changes are made: 

- Using `table, next` in a for loop on an array is not more efficient. See Crazyman32's [article](https://medium.com/@RBX_Crazyman32/don-t-use-next-instead-of-pairs-6758bb19de61#.hcmcpxkkq) on it. Please change to the numeral for loop (for efficiency) or just use the `pairs()` method, as that is more clear to beginners. 
- Please avoid (in most cases) putting more than one command per a line, especially on property tables when using `make`. This is for readability. Please change the few cases where stuff got moved to more than one line. 
- Move the `elseif type(key) == "function"` code in the make to be after the type check for properties, so the speed of existing make code is relatively not affected by performance. 
    - This is a nitpick, just a personal preference. 
- Although I'd prefer that you use `CamelCase` as long as the exposed API is consistent, it's fine. Just thought I'd note that, you don't need to make any change.

I'm sorry I'm being a stickler for these things, they're unwritten, but I want to maintain a certainly level of quality in this codebase. These rules are-as always-up for debate. 

I'm sorry for the delay, I'll get to your other pull requests soon.

Narrev:
I changed Make to use a generic 'for' loop instead. If I were to move around the checks, look at the differences in the number of checks:
```lua
-- New
for Index, Value in next, Values do
	if type(Index) == "string" and Index ~= "Parent" then
		Instance[Index] = Value
	elseif type(Index) == "number" then
		Value.Parent = Instance
	elseif type(Value) == "function" then
		Instance[Index]:connect(Debounce(Value))
	end
end
```
```lua
-- old
for Index, Value in next, Values do
	if type(Index) == "number" then
		Value.Parent = Instance
	elseif type(Value) == "function" then
		Instance[Index]:connect(Debounce(Value))
	elseif Index ~= "Parent" then
		Instance[Index] = Value
	end
end
```
Instance[Index] = Value would have 2 checks, which is 1 less than the old function.
Value.Parent = Instance would have 3 checks, which is 2 more than the old function
Instance[Index]:connect(Debounce(Value)) would have 4 checks, which is 2 more than the old function.

Quenty:
Whoah, why does it modify Parent last?

Narrev:
Because if you don't modify Parent last, setting CFrame of a brick can be undone

Quenty:
Then instead of doing a variable amount of 10-20 if statements + function calls, just allow it to set parent, just like CFrame, and then set Parent AFTER CFrame, assuming you set CFrame.

Don't over complicate what is a simple helper method. 

```lua
if Values.CFrame then
    Instance.CFrame = Values.CFrame
        if Values.Parent then
            Instance.Parent = Parent
    end
end
```

However, it looks like these methods are getting _really_ heavy, to the point that performance may become an issue.

I suggest that you move functionality into a second function, since using Nevermore shouldn't create a ton of overhead.

Narrev:
Okay, I moved the checks around. I think it works great now. These methods aren't heavy. We only added one extra check to Make and one extra check to modify that allows connections.

Quenty:
> We only added one extra check to Make and one extra check to modify that allows connections

The thing is that those checks are in loops, so in some cases, they get called hundreds of times. I think it'll still be fine. 

I'll look at this branch and merge it relatively soon, assuming all is well.

Narrev:
I don't think it will make much of a difference at all. It is hardly slower than it was before I made it possible to cut 75 lines out of a script

PR #38, opened 2016-04-06
Improve os.date efficiency by another 12%

Narrev:
Uses [these algorithms](http://howardhinnant.github.io/date_algorithms.html#weekday_from_days)


Quenty:
Looking good Narrev.

PR #40, opened 2016-04-10
Error if TeamCreate is enabled

Narrev:
When NetworkClient is open, HTTP requests can only be performed on an actual server, and not from the command bar (or plugins?)


Quenty:
Merge pull request

PR #43, opened 2016-04-21
Update Signal.lua

Narrev:


Modules/Events/Signal.lua
```diff
@@ -25,43 +33,37 @@
...
+		return unpack(BindData)
 	end
 
-	function sig:Destroy()
-		mSignaler:Destroy()
-		mArgData      = nil
-		mArgDataCount = nil
-		mSignaler     = nil
+	function Signal:Destroy()
+		BindData = BindableEvent:Destroy()
```
Quenty: Probably better to just set these values to nil on a separate line to clarify what is happening here. 

Note that :Destroy() is a bit of a nasty method because the connector should be the one disconnecting / handling event connections to avoid keep-alive memory leaks.

Narrev:
My mistake. I thought I read that Destroy()ing a BindableEvent Object automatically disconnects any connections associated.

Quenty:
Destroying a BindableEvent does automatically disconnect it, I meant for Signal() itself, it can be misleading in terms of who controls GC

Narrev:
Updated

PR #45, opened 2016-06-10
Add Loading Screen!

Narrev:
Makes installation a lot cooler


Quenty:
Will consider in conjunction with #49, but will not be minimized nicely.

PR #46, opened 2016-06-20
Deprecate qSystems and move Folders

Narrev:


Quenty:
Looking to merge this. Can you summarize changes and note backwards compatability state?

Narrev:
I have added a few new functions, but the majority of things done in this patch is just moving things around. I deprecated qSystems, and moved its functions into player functions, math functions, and helper functions. Using qSystems will still work, but it is officially deprecated. All of the qSystems functions are pulled from elsewhere. 

I think this version of Nevermore is a bit more of a healthy balance between having libraries and having some isolated functions that don't make very much sense in a library. I reorganized a lot of the module locations so that their position in Folders makes sense, but I also didn't want to make it too big to approve. 

I did test this, and everything should be backwards compatible. That's what I had in mind as I was making the changes.

PR #49, opened 2016-07-09
Make Installation code shorter

Narrev:


Quenty:
Doesn't appear to work when loadstring is disabled. Can you verify?

Quenty:
It would also be a good idea to verify the source-code is what we want to execute with a given hash (man-in-the-middle attacks), using some sort of hashing. See https://packagecontrol.io/installation for an example (Although Lua has no nice hashing library at this point to import).

Narrev:
The code runs from your command bar. It doesn't matter whether or not loadstring is enabled because the installation code doesn't run in a script.

Quenty:
Nice. Can you please

* Write this code fully into another script
* Minimize that code and put it into the readme.md

This should mean allow for future loading code to be easily maintained.

Narrev:
It is 4 lines. It just turns on HttpEnabled temporarily and then runs the installation script from the GitHub website. It doesn't need to be maintained. Also, the only way it could be minimized is to remove newlines. I think it looks nicer to just copy and paste 4 lines than to try to oddly copy and paste a horizontal scroller bar. 

If you don't like that idea, maybe we should set up something a little more flexible like [this](http://sleitnick.github.io/RobloxTweenAPI/)? That would allow for clicking on a button to copy to clipboard. Otherwise, I don't think there is any issue with this pull request.

Quenty:
I think it's important to try to maintain code readability, although in this case, if we don't add hashing to the source code validation, then this should be perfectly fine. I think moving it into one line makes sense though, as ROBLOX studio does not support multiple lines in the command bar. 

I'd love to try to setup something like the link, or perhaps something that even lets you select libraries to install with Nevermore.

However, it would also be nice to move Nevermore's libraries out towards other github repositories and have Nevermore act more like a package manager.

Narrev:
I think it would be good to consider the possibility of creating a Nevermore plugin. That would eliminate the need for installation code. 

I also don't think there are that many other worthy Github repositories out there. I would love for you to point some out, but most people that open-source just publish to Roblox models.

If you don't require a module, it doesn't take up anything, right? Having modules sit there doesn't really do anything, right?

This is why I think NevermoreEngine should be called Nevermore in the future.

Also, Roblox does support multiple lines in the command bar. This is what it looks like when you paste the 4 lines of code:

cmd bar

VoxelDavid:
I gotta say, I like the idea of a Nevermore plugin quite a bit. It could be used for installing Nevermore in the first place, with checkboxes to select which of the base modules you want to install. Maybe it could even manage your installed modules? It could give you a nice interface to delete any of the ones you don't want anymore, including your own (though that's easy enough to do by hand in Studio).

And as a user, I would advise against splitting up all of the modules into their own repos. I've had experience with a library that has tons of different modules split up between repos and it was very tedious and annoying to track down all the ones I wanted for my project. If you do go with this approach I would definitely recommend a plugin that collects all of your base modules so the user can install them without having to track down every repo.

Narrev:
I think ultimately your problem should be addressed with more documentation. Petition Quenty to add documentation whenever he adds something

VoxelDavid:
That's a good point. More documentation would definitely be helpful, particularly in introducing the user to all the modules and what they do. I don't actually know much about any of the modules because I don't have the inclination to read the source, and the docs don't explain much (though it's also been a while since I've look at the modules, maybe that's changed).

I think Nevermore would be great as a solo package manager, but I'm concerned about how easily accessible all the modules would be. If they're in their own repos there would need to be clear documentation so users actually know they exist, and how to find them (maybe a repo wiki page listing them all? That might get forgotten easily. Web interface that pulls them all in, maybe? Kind of like an [npm](http://npmjs.org) sorta thing.)

Quenty:
I want to turn Nevermore into a solo package manager similar to npm, which I've discussed with Narrev before.

NevermoreEngine is definitely misleading, and I'll create an issue to rename in the future (See #52). Certain modules should definitely be culled in the future for reliability.

Design is certainly a concern here though. I'll be thinking about this in the future.

Narrev:
You know how Nevermore has a list of features on the help page? I think the plug-in should list those, and when you click on one it expands and shows various libraries you can check a box to install. It also will automatically check boxes that are needed to run that library. If you try to uncheck a library, it will warn you of the other modules that rely on it to run.

Narrev:
Things like "Nevermore Chat" can be checked, and all the modules listed under it will be checked

PR #50, opened 2016-08-20
Procedurally Generate GetRemoteEvent+Function

Narrev:
This new version is shorter, more efficient, has a more flexible syntax, supports managing more instances other than just RemoteEvents and RemoteFunctions, and supports Appendages! More information about Appendages here: https://github.com/Narrev/Nevermore
Also, completely backwards compatible!
And it comes with a built-in (generated) Folder manager for use with the NevermoreResources folder!

### DIFF
--- a/App/NevermoreEngine.lua
+++ b/App/NevermoreEngine.lua
@@ -89,23 +89,12 @@ function self:__index(index) -- Using several strings for the same method (e.g. 
 	local originalIndex = index
 	local index = string.gsub(index, "^Get", "")
 	local Class = Classes[index] or index
-	local Appendage = script:FindFirstChild(index)
-	local Function
+	local Table = {}
+	local Folder = self:Folder(Class .. "s")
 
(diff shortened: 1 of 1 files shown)


Quenty:
Looks pretty good. Will be looking into this implementation more before merging.

Narrev:
I realized that I didn't change the FolderName in config back to "Nevermore". I have been using "Modules" because it's more intuitive, but I recognize that the Nevermore setup code creates a folder named "Nevermore". 

This may have prevented this new version from working for you if you didn't catch this.

It now should be fully capable of running your codebase.

Quenty:
Looking very good. Is "extract" intentionally lower case? If not, we should change it. A style guide on the wiki would also be good.

Narrev:
Yes, `extract` is intentionally lowercase. It is meant to mimic the syntax for `select`

Narrev:
Why hasn't this been implemented yet?

Narrev:
Outdated

PR #54, opened 2016-10-10
Finish os.lua

Narrev:
I clearly check on this code every so often, because I wrote it, and it needs to be functioning optimally. This pull request makes it so the module only generates the string tables a single time, and doesn't generate new tables for each use.

### DIFF
--- a/Modules/Utility/os.lua
+++ b/Modules/Utility/os.lua
@@ -1,5 +1,9 @@
 -- @author Narrev
 
+-- Necessary string tables
+local dayNames = {[0] = "Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"}
+local months = {"January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"}
+
 local function date(formatString, unix)
 	--- Allows you to use os.date in RobloxLua!
 	--		date ([format [, time]])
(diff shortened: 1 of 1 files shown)


Modules/Utility/os.lua
```diff
@@ -1,5 +1,14 @@
...
+-- Necessary string tables
+local dayNames = {[0] = "Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"}
+local months = {"January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December"}
+
+-- Localize functions
+local floor, sub, find, gsub, format = math.floor, string.sub, string.find, string.gsub, string.format
+local time = os.time
+local type, wait, tick, tonumber, assert = type, wait, tick, tonumber, assert
+
 local function date(formatString, unix)
```
Quenty: Might make sense to rename `unix` to be `unixTime` (we probably don't want to refactor the whole class to match the wiki standards at this point, until those are better defined) (someday?).

I would

Modules/Utility/os.lua
```diff
@@ -35,7 +41,7 @@ local function date(formatString, unix)
 			local UTC
 			formatString, UTC = gsub(formatString, "^!", "") -- If formatString begins in '!', use os.time()
 			assert(UTC == 0 or not unix, "Cannot determine time to format for os.date. Use either an \"!\" at the beginning of the string or pass a time parameter")
-			unix = UTC == 1 and os.time() or unix
+			unix = UTC == 1 and time() or unix
```
Quenty: I would prefer this at the top of the date function which helps clarify behavior of arguments if you're reading this code (this is pretty standard).

Thoughts

Quenty:
I'm willing to merge now, but will wait for responses on those two things.

Also, something about the chained gsubs bothers me (lots of copied and pasted code?), but it'll be significant refactoring to change this.

Narrev:
I did not copy and paste the code. This library isn't like other libraries because this should be built-in. I've included instructions in the README, as well as the fact that you can look up how to use os.date from a variety of sources. My module is designed to be functional and fast, even at the expense of feeling uncomfortable about a long chain of gsubs. This isn't the kind of library you should need to read through. You should be able to load it and start using it without reading the module.

PR #62, opened 2017-05-06
Refactor Nevermore

Narrev:
This is the latest iteration of the Validark rewrite of Nevermore.

Differences:
* Focuses more on optimization but not at the cost of functionality. While this version doesn't include localizing all the variables at the beginning (for formatting's sake), making Nevermore run faster was definitely a major reason for rewriting it.
* Functions are procedurally generated (upon initial use only), so Nevermore will generate functions for any Folder or Method the developer's heart desires (see code below)
* Added ability to put the word "Local" after "Get" to make the function access local resources, e.g. `local Jump = Nevermore:GetLocalAnimation("Jump")`
* Intelligent error system is universally applied
* Made descendants of any Object with Server in its name stored in ServerStorage (LocalStorage on the Server, which is also accessible via the LoadLibrary function)
* Made Modules under an object called "Server" able to share a name with another Module, but on the Server the former is the only accessible one (on the client, the server version doesn't exist, and thus only the client version of the module of the same name is accessible). This is useful because you can have two versions of a module, 1 for the client, and 1 for the server. Totally optional though.
* Made Nevermore generate `GetFolder` by default (it's used internally). If you believe it should be removed in favor of just calling GetFirstChild, that is an idea we should discuss
* Made it slightly more lengthy due to the longer "LibraryCache Assembly" as I used to call it. While it has more going on, it is much more faster than the old version.
* Removed DebugMode because its results are more or less built-in now
* Made all functions support calling by either ":" or "." (hybrid syntax)
* UPDATE: Functions are functions again. Originally we switched to metatable + calling because devSparkle wanted to see if there would be storage benefits to sharing a single function. It was a little slower due to that, but Quenty indicated his focus on speed so we are back to functions

```lua
local Replicator = Nevermore:GetRemoteEvent("Replicator")
local BindableEvent = Nevermore:GetLocalBindableEvent("Standard")
```

Let me know if you have any questions or concerns.
Validark

### DIFF
--- a/App/NevermoreEngine.lua
+++ b/App/NevermoreEngine.lua
@@ -195,9 +195,9 @@ local LibraryCache = {}
 function Nevermore.LoadLibrary(self, Name) -- Custom Require function
 	Name = self ~= Nevermore and self or Name
 	local Library = LibraryCache[Name]
-	if not Library then
+	if Library == nil then
 		Library = require(GetModule(Name))
-		LibraryCache[Name] = Library
+		LibraryCache[Name] = Library or false
 	end
 	return Library
 end
--- a/App/NevermoreEngine.lua
+++ b/App/NevermoreEngine.lua
@@ -1,5 +1,5 @@
 -- @author Validark
--- @readme https://github.com/Quenty/NevermoreEngine
+-- @readme https://github.com/NevermoreFramework/Nevermore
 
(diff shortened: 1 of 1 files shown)


Narrev:
Bump

Narrev:
Good luck, Quenty

PR #64, opened 2017-07-06
Fix TopRight, BottomLeft calculations

Narrev:
The parameters of the Vector2 need to be switched.


Narrev:
The order for Vector2s is (X, Y) and not the other way around

PR #70, opened 2017-12-10
Make Installing nearly instantaneous

Validark:
- Eliminate need for python script via universal Github Installer
- Make Install script change HttpEnabled back to what it was

Test code:
```lua
local h = game:GetService("HttpService") local e = h.HttpEnabled h.HttpEnabled = true loadstring(h:GetAsync("https://raw.githubusercontent.com/Validark/NevermoreEngine/patch-3/Install.lua"))(e)
```


Install.lua
```diff
@@ -1,175 +1,275 @@
...
+	-- Installs Link into Parent
+	Link = Link:gsub("/$", "")
+
+	-- Extract Link Data
+	local Organization, Repository, Tree, ScriptName, ScriptClass
+	local Website, Directory = Link:match("^(https://[raw%.]*github[usercontent]*%.com/)(.+)")
+	Organization, Directory = (Directory or Link):match("^/?([%w-_%.]+)/?(.*)")
+	Repository, Directory = Directory:match("^([%w-_%.]+)/?(.*)")
+
+	if Website == "https://raw.githubusercontent.com/" then
```
Quenty: Code clarify:

Should break this if-statement up into several methods.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+		if Directory then
```
Quenty: Can be a method called `Github:ExtractScriptData(Directory)` which returns a table with ScriptName/ScriptClass. Make it much more clear what you're doing.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+			Tree, Directory = Directory:match("^([^/]+)/(.+)")
+			if Directory then
+				ScriptName, ScriptClass = Directory:match("([%w-_%%]+)%.?(%l*)%.lua$")
+			end
+		end		
+	elseif Directory then
+		local a, b = Directory:find("^[tb][rl][eo][eb]/[^/]+")
```
Quenty: Makes it hard to tell blob/tree is being extracted beforehand. Maybe extract `(%w+)/([^/]+)` and check the first word if it's tree/blob, and the second for the match using `ExtractScriptData` method suggested instead of repeated code.  

In reality this is to grab scriptdata from a directory, so probably break this up too.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+			if Directory == "" then Directory = nil end
+			if Directory and Link:find("blob") then
+				ScriptName, ScriptClass = Directory:match("([%w-_%%]+)%.?(%l*)%.lua$")
+			end
+		else
+			Directory = nil
+		end
+	end
+
+	if not Website then Website = "https://github.com/" end
```
Quenty: Can replace with `Website = Website or "https://github.com"` pattern, otherwise probably tab correctly.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+	Directory = Directory and ("/" .. Directory):gsub("^//", "/") or ""
+
+	-- Threads
+	local Routines = RoutineList or {false}
```
Quenty: Method shouldn't do both extraction and task scheduling like this, interloping with global variables

Quenty: Instead: split up into multiple variables/methods

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+	local Value = #Routines + 1
+	Routines[Value] = false
+
+	local Garbage
+
+	if ScriptName then
+		local Script = Instance.new(ScriptTypes[ScriptClass or "mod"])
```
Quenty: This should be in a method like `CreateScript()`

Install.lua
```diff
@@ -1,175 +1,275 @@
...
-MakeScript(ReplicatedStorage, "App/NevermoreEngine.lua")
+		if MainExists then Directory = Directory + 1 end -- :gsub("[^/]+$", "", 1) end			
+		local Count = 0
+
+		local function LocateFolder(FolderName)
+			Count = Count + 1
+			if Count > Directory then
+				Directory = Directory + 1
+				if (Parent and Parent.Name) ~= FolderName and "Modules" ~= FolderName then
+					if FolderName == "App" and Link:find("/Quenty/NevermoreEngine/", 1, true) then FolderName = "ReplicatedStorage" end
```
Quenty: Note for the source of this file, should avoid hardcoding this stuff in. OK in an installer. Bit weird in a generic installer module.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
+		DataSources[Object] = Link
+		return Object
+	end
 end
 
-HttpService.HttpEnabled = WasHttpEnabled
-ProgressText.Text = "Done!"
-done = true
+GitHub:Install("/Quenty/NevermoreEngine", game:GetService("ServerScriptService")).Name = "Nevermore"
+HttpService.HttpEnabled = ...
```
Quenty: Should leave a comment why we're doing this/usecase, incase someone tries to install this the other way.

Install.lua
```diff
@@ -1,175 +1,275 @@
...
-local Paths = {}
-for ScriptPath in (LoadURL("Modules/ModuleList.txt")):gmatch("[^\r\n]+") do
-	Paths[#Paths+1] = ScriptPath
-	print""
+		for a = 2, ScriptCount do
+			local Link = Scripts[a]
+			local ScriptName, ScriptClass = Link:match("([%w-_%%]+)%.?(%l*)%.lua$")
+			local Script = Instance.new(ScriptTypes[ScriptClass or TypesSpecified and "" or "mod"], Parent)
+			Script.Name = ScriptName:gsub("Library$", "", 1):gsub("%%(%x%x)", UrlDecode)
+			coroutine.resume(coroutine.create(GiveSourceToScript), "https://raw.githubusercontent.com" .. Link:gsub("(/[^/]+/[^/]+/)blob/", "%1", 1), Script)
```
Quenty: coroutine passing like this is hard to follow, need to be careful in certain cases. DownloadScriptSource maybe better name? Concatenation of link here hard to follow.

Quenty:
Oh yeah, also please remove the python installer and the install list! We won't need it anymore.

Validark:
- Detects previous install
- Updates existing versions.

Quenty:
Not perfectly clean, but we'll take this. Nice job!

Quenty:
Before we can merge, you'll need to merge in the latest version than `git rm ModuleList.txt` and the `CompileBuilder.py`. Sorry!

Validark:
You'll have to delete the ModuleList.txt yourself.

Validark:
Ready for merging.

Quenty:
Merged! Thank you!

PR #75, opened 2017-12-19
Bezier 2.0

### LINKED ISSUE #69: BezierFactory.lua is based on deprecated files
Deprecated version:
https://gist.github.com/gre/1926947#file-keyspline-js

Here is the recommended version:
https://github.com/gre/bezier-easing

Validark:
- Rename it `Bezier`
- Give `Bezier` the standard syntax `Bezier.new`
- Update `Bezier` according to https://github.com/gre/bezier-easing
- Make `Bezier` compatible with standard easing functions that have parameters (b, c, d)
- Make `Bezier` Great Again
- Make `Bezier` use `2` as second parameter in `error` so that the `error` points to the function that called `Bezier` improperly
- Make `Bezier` backwards compatible
Fixes #69


Validark:
```lua
-- Taken from material.io motion guidelines
local Sharp = Bezier.new(0.4, 0, 0.6, 1)
local Standard = Bezier.new(0.4, 0, 0.2, 1)
local Acceleration = Bezier.new(0.4, 0, 1, 1)
local Deceleration = Bezier.new(0, 0, 0.2, 1)
```

Quenty:
I can't merge these changes. LDoc doesn't compile over it, and the code is generally unreadable. 

Please avoid single letter variables at the top of the scope, and keep the code simple.

Also, @see is to point towards a module, but in this case, we don't want any author tags since authors are tracked by github.

Validark:
The code is simple. Oh my gosh...

Validark:
"generally unreadable"

How... It is so concise and simple

Validark:
Just because I have a variable instead of a function call to `A` does not make it less readable.

Quenty:
You're right. Sorry, I came off as really abrasive. I want your changes.

To clarify, the list of single variable "redundant values" makes it hard for anyone to understand what is happening in the code. Preferably these calculations are put into functions, and/or named appropriately. Alternatively, if they're used within a single scope (instead of across some 70ish lines of code), they're fine as single variables.

https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R37



Lines such as this should have variables in `UPPER_CASE` to show they're constants.

https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R21

Lines where `if ... then break end` should be broken up into multiple lines. However, this makes flow still unclear. It's better to move these into a separate function call and immediately return the result.

https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R96

Naming loops such as these:
https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R77
https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R85

Can help make code more clear. For example, `calculateSlope()` could be good.

Comments such as these:
https://github.com/Quenty/NevermoreEngine/pull/75/commits/0bdf3c5c31a6286644d1cfe91e5170548c6a6ec3#diff-b2d653656e1a0d81273ec44ff0e31b63R85

Can be replaced with a function named `estimateTime()` which most modularize the code and breaks up the code. 

Does that help clarify things?

Validark:
I don't want to move things into separate functions. I don't think that what matters most is whether a well-named function could make someone understand the math behind it. There is no way to understand it without research anyway. It isn't an explanation for how Beziers work, it is a module you can plug and play to do one thing: generate Bezier functions. The single letter variables don't have a proper name because there is no proper name to give them. It's just math I chose to place outside the function so that the function doesn't end up doing the same operations over and over with each function call (which typically happens hundreds or thousands of times). I just don't agree with your calling it 70 lines and saying it does too many things. It is spaced out so it can be more readable. It only makes a single Bezier function. Frankly, I don't see a need for `GetSlope`. Nobody needs `GetSlope`, they just want a Bezier Factory, preferably a speedy one without tons of unnecessary abstraction nobody is going to look at.

I'd rather put in more comments instead of using function names as documentation.

Validark:
I don't agree with replacing comments with more functions

Quenty:
Unfortunately, at this time I won't be merging this request. I understand why you made the change, but this isn't the direction I'd like to take Nevermore!

Thank you for your time!

PR #89, opened 2018-03-10
Memoization as OOP instead of functional

Quenty:


Quenty:
Wait. This doesn't work. `__call` can't yield across boundaries.

Quenty:
This is now fixed.

App/Nevermore.lua
```diff
@@ -15,49 +15,46 @@ assert(script.Name == "Nevermore", "Invalid script name. For Nevermore to work c
...
-			return result
-		end
+local AsyncMemoizer = {}
+AsyncMemoizer.ClassName = "AsyncMemoizer"
+AsyncMemoizer.__index = AsyncMemoizer
 
-		bindableEvent.Event:Wait()
-		return result
-	end
+--- Caches single argument of an asyncronious function
```
blobbyblob: This could benefit from improved documentation.

```lua
--@param func A function which takes one argument and returns one result. This function will only be called once for each unique argument.
--@return A function which, when called, invokes the function & caches its result.
```

App/Nevermore.lua
```diff
@@ -182,20 +179,20 @@ local lib = {}
 -- @function LoadLibrary
 -- @tparam Variant LibraryName Can either be a ModuleScript or string
```
blobbyblob: Out of curiosity, what is the implication of loading the same library twice, once with its name, and once with the ModuleScript reference?

App/Nevermore.lua
```diff
@@ -15,49 +15,46 @@ assert(script.Name == "Nevermore", "Invalid script name. For Nevermore to work c
...
+	end
 end
 
---- Caches single argument, single output only
-local function asyncCache(func)
-	assert(type(func) == "function", "Error: func must be a userdata")
-
-	local cache = {}
-	local retrieving = {}
+function AsyncMemoizer:_pend(arg)
```
blobbyblob: Maybe an assert would be good here to ensure `arg` is not nil.

```lua
assert(arg ~= nil, "AsyncMemoizer does not support nil arguments");
```

Though probably from a usability standpoint, you'd want this to not call itself an AsyncMemoizer, but "LoadLibrary", "GetRemoteEvent", or "GetRemoteFunction". Dunno, food for thought. It seems nasty to throw that in as another parameter for the constructor, but perhaps the constructor could return both the object, and the function? Or the function could be a (loosely-enforced) read-only parameter of the object?

```lua
local loadLibraryObj = AsyncMemoizer.new(debugLoading(getLibraryLoader(libCache)))
loadLibraryObj.DebugName = "LoadLibrary"
lib.LoadLibrary = loadLibraryObj.Call
```

App/Nevermore.lua
Quenty: It should function exactly the same. Roblox handles the caching.

App/Nevermore.lua
Quenty: Agreed!

App/Nevermore.lua
Quenty: This is a great idea. Personally I think the `GetRemoteEvent` and `GetRemoteFunction` need to go into a helper library (and because of how Nevermore works, this will not break backwards compatability). 

Construct shouldn't probably return two things? I'm not sure. Maybe an :apply() method. Would have to look back at the context of the code.

PR #95, opened 2018-08-22
Update Table.lua

howmanysmall:
Removed some useless code (that has no benefit) and optimized.

### DIFF
--- a/Modules/Utility/Table.lua
+++ b/Modules/Utility/Table.lua
@@ -36,21 +36,9 @@ end
 function lib.ToList(tab)
 	local list = {}
 	for _, item in pairs(tab) do
-		table.insert(list, item)
+		list[#list+1] = item
 	end
 	return list
-end
-
---- Counts the number of items in `tab`.
--- Useful since `__len` on table in Lua 5.2 returns just the array length.
--- @tparam table tab Table to count
--- @treturn number count
-function lib.Count(tab)
-	local count = 0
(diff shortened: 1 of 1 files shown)


Modules/Utility/Table.lua
```diff
@@ -36,23 +36,11 @@ end
...
+		list[#list+1] = item
 	end
 	return list
 end
 
---- Counts the number of items in `tab`.
--- Useful since `__len` on table in Lua 5.2 returns just the array length.
--- @tparam table tab Table to count
--- @treturn number count
-function lib.Count(tab)
```
Quenty: `lib.Count` works on  non-array components as the document specifies. We can't remove this method.

Modules/Utility/Table.lua
```diff
@@ -167,10 +155,10 @@ end
 function lib.ReadOnly(table)
 	return setmetatable(table, {
 		__index = function(self, index)
```
Quenty: This is good!

Modules/Utility/Table.lua
```diff
@@ -36,23 +36,11 @@ end
 function lib.ToList(tab)
 	local list = {}
 	for _, item in pairs(tab) do
-		table.insert(list, item)
```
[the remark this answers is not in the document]
Quenty: Personally I prefer table.insert's explicit list, but technically list[#list +1] is faster. I think table.insert reads nicer.

PR #99, opened 2018-10-30
Update FastSpawn.lua

howmanysmall:


Modules/Utility/FastSpawn.lua
```diff
@@ -1,16 +1,19 @@
 --- An expensive way to spawn a function. However, unlike spawn(), it executes on the same frame, and
 -- unlike coroutines, does not obscure errors
 -- @module fastSpawn
+local inst = Instance.new
```
Quenty: This performance improvement isn't worth the loss in readability! Sorry!

Modules/Utility/FastSpawn.lua
```diff
@@ -1,16 +1,19 @@
...
-	assert(type(func) == "function")
-
-	local args = {...}
-	local bindable = Instance.new("BindableEvent")
-	bindable.Event:Connect(function()
-		func(unpack(args))
-	end)
-
+	local bindable = inst("BindableEvent")
+	if ... ~= nil then
```
Quenty: I don't thinks this performance improvement is really significant. Creating a new bindable and destroying it is magnitudes and magnitudes slower than the construction of the small table in the case of empty arguments.

Additionally, if statements may be slow. 

I can't approve this pull request as is, please confirm what this change is intending to do.

PR #100, opened 2018-10-30
Fix invalid module in Binder.lua

JohnnyMorganz:
The current version of Binder requires “fastSpawn”, however the name has been capitalised in the repository to “FastSpawn”. This just fixes that

I don’t know if you want to change the variable name to FastSpawn - in line with Maid being capitalised.


Quenty:
Thank you for your contribution. fastSpawn should be lowercase, it's a function. Maid is a class, so it's uppercase. 

This follows Roblox's style guidelines. I'm going to go ahead and rename the file.

Quenty:
Thanks! This should be fixed now!

PR #112, opened 2019-01-21
Update init.lua

howmanysmall:
`,` over `..`

### DIFF
--- a/Modules/InputImageLibrary/init.lua
+++ b/Modules/InputImageLibrary/init.lua
@@ -86,7 +86,7 @@ function InputImageLibrary:_pickSheet(index, preferredStyle, preferredPlatform)
 		end
 	end
 
-	warn("[InputImageLibrary] - Unable to find sprite for " .. tostring(index) .. " type " .. typeof(index))
+	warn("[InputImageLibrary] - Unable to find sprite for", tostring(index), "type", typeof(index))
 
 	return nil
 end


Modules/InputImageLibrary/init.lua
```diff
@@ -86,7 +86,7 @@ function InputImageLibrary:_pickSheet(index, preferredStyle, preferredPlatform)
 		end
 	end
 
-	warn("[InputImageLibrary] - Unable to find sprite for " .. tostring(index) .. " type " .. typeof(index))
+	warn("[InputImageLibrary] - Unable to find sprite for", tostring(index), "type", typeof(index))
```
Quenty: Might as well use a format string here

Modules/InputImageLibrary/init.lua
```diff
@@ -101,4 +101,4 @@ function InputImageLibrary:_getImageInstance(instanceType, index, preferredStyle
 	return nil
 end
 
-return InputImageLibrary.new(script:WaitForChild("Spritesheets"))
\ No newline at end of file
+return InputImageLibrary.new(script:WaitForChild("Spritesheets"))
```
Quenty: We should decide upon a standard, return or no return, for code

Quenty: Approved other PRs with changes to end line.

PR #115, opened 2019-01-24
Switch documentation to makedocs

Quenty:


Quenty:
Closing since we switched to Moonwave.

PR #121, opened 2019-03-12
Update Table.lua

howmanysmall:
Add FastRemove as a function. Can be used in place of `table.remove` where you don't need to return a value.


Quenty:
Benchmarks on this vs. Table.Remove?

howmanysmall:
image
image
image

PR #123, opened 2019-07-01
Update init.lua

colbert2677:
**Specific update:** private method replicateRepository

I'm not quite sure if it's a misunderstanding of mine, but I don't believe the original code actually removed client-only libraries. It simply got names of client-only libraries and placed them into a toRemove table, then set an indice in the toRemove table with the library's name to nil. Perhaps that was meant to set the indice in libCache to nil to actually remove it from the cache?


Quenty:
Thank you!

PR #128, opened 2019-09-12
Update CollectionServiceUtils.lua

howmanysmall:
Added `findNearestPartWithTag`. Used for searching around a given BasePart to find what part with the given tag is nearest to it.

Sorry if the code is inconsistent with your style, it's my best attempt at mocking the preexisting code.


howmanysmall:
Yeah, that would be better. Thanks for pointing that out, I'll make an edit.

PR #136, opened 2019-11-14
small typo on datastore documentation for datastore

mobyboy:
fixed, just felt like it was needed to be pointed out.


Quenty:
thank you!

PR #140, opened 2020-02-11
Update fastSpawn.lua

Rerumu:
Function can be simplified. Use of a single BindableEvent greatly reduces overhead.

### DIFF
--- a/Modules/Shared/Utility/fastSpawn.lua
+++ b/Modules/Shared/Utility/fastSpawn.lua
@@ -1,18 +1,12 @@
 --- An expensive way to spawn a function. However, unlike spawn(), it executes on the same frame, and
 -- unlike coroutines, does not obscure errors
 -- @module fastSpawn
+local bindable = Instance.new('BindableEvent')
+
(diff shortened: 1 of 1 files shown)


howmanysmall:
See: https://github.com/roblox-ts/roblox-ts/issues/668

howmanysmall:
For one with a little less overhead you can do this:

```Lua
local function FastSpawnV2(Function, ...)
	local Length = select("#", ...)
	if Length > 0 then --if ... ~= nil then
		local BindableEvent = Instance.new("BindableEvent")
		local Arguments = {...}
		BindableEvent.Event:Connect(function()
			Function(table.unpack(Arguments, 1, Length))
		end)

		BindableEvent:Fire()
		BindableEvent:Destroy()
	else
		local BindableEvent = Instance.new("BindableEvent")
		BindableEvent.Event:Connect(Function)
		BindableEvent:Fire()
		BindableEvent:Destroy()
	end
end
```

But it's not really worth it.

Quenty:
Huh. So reusing bindables actually can create an issue. Weird note!

Issue #139, opened 2020-02-02
Add varargs to maid:GiveTask to support adding multiple tasks at once


sayhisam1:
Currently, the procedure to add many items to maid is to do a repeated call to `GiveTask`, as so:

`
    local a = Instance.new("Part")
    local b = Instance.new("Part")
    maid:GiveTask(a)
    maid:GiveTask(b)
`

I propose adding varargs support to :GiveTask, so that we can simply do:
`
local a = Instance.new("Part")
local b = Instance.new("Part")
maid:GiveTask(a,b)
`
instead, which is a lot cleaner as it fits in a single line.

devSparkle:
For the sake of long-term support, it would probably be wiser to add the ability to provide multiple objects to clean _as a table._ If we were to add varargs, this could prevent the addition of custom arguments to the `GiveTask` method later on.

sayhisam1:
> 
> 
> For the sake of long-term support, it would probably be wiser to add the ability to provide multiple objects to clean _as a table._ If we were to add varargs, this could prevent the addition of custom arguments to the `GiveTask` method later on.

It would probably be best to have a new method altogether, then - say `maid:GiveTasks(...)`

Quenty:
I'd be more comfortable with :GiveTasks(), but I'm not entirely sure this is something we want to do. It's really nice to keep the :GiveTask() separate.

PR #143, opened 2020-03-28
maid:GiveTask varargs

05st:
#139

### DIFF
--- a/Modules/Shared/Events/Maid.lua
+++ b/Modules/Shared/Events/Maid.lua
@@ -56,19 +56,24 @@ function Maid:__newindex(index, newTask)
 end
 
 --- Same as indexing, but uses an incremented number as a key.
--- @param task An item to clean
--- @treturn number taskId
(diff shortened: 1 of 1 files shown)


Modules/Shared/Events/Maid.lua
```diff
@@ -56,19 +56,24 @@ function Maid:__newindex(index, newTask)
...
-function Maid:GiveTask(task)
-	assert(task, "Task cannot be false or nil")
+-- @param tasks Item(s) to clean
+-- @treturn table taskId(s)
+function Maid:GiveTask(...)
+	local taskIds = {}
 
-	local taskId = #self._tasks+1
-	self[taskId] = task
+	for task in pairs({...}) do
```
Halalaluyafail3, line 64: This looks very wrong, this is iterating over the number 1 to the amount of arguments, in any order. 'task' will be a number. The values of the arguments are unused here. Was something like 'for _,task in ipairs{...}' intended?

PR #144, opened 2020-05-12
Module Loader doesn't load modules

kieranz:
The ModuleLoader tries to load modules from Nevermore, however, that folder doesn't exist.


Quenty:
Closing, installer was fixed recently.

PR #152, opened 2020-07-14
Update AddModule example in loader README

swenzje:


Quenty:
Can you add a separate example, instead of removing the other example? Otherwise, looks good! 

Also, there's no need to :WaitForChild() in ReplicatedStorage!

PR #153, opened 2020-07-16
Update MasterClock.lua

howmanysmall:
Remove random spaces.


Quenty:
Thank you for breaking up these pull requests into multiple small reviewable ones!

PR #155, opened 2020-07-18
Upgrade Installer

Validark:
Finally got around to refactoring the Installer. It now works using Promises and GitHub's official API, so it shouldn't break anymore :)

(the old version manually scraped GitHub webpages, which led to problems)

### DIFF
--- a/Install.lua
+++ b/Install.lua
@@ -1,328 +1,776 @@
--- Universal Github Installer
--- @see Validark
--- function GitHub:Install(Link, Parent)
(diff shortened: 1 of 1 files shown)


Validark:
Since this isn't a real library, could this be excluded from the lint check?

IntelligentCappy:
I still think this is relevant. Couldn't install with the installer that's included with the base branch due to an infinite yield with a repeat until loop on line 277. This one worked fine.

Quenty:
You could add --luacheck: ignore at the top, but I'd prefer someone fix the linting. Either way is fine.

Issue #147, opened 2020-06-03
Add support for alternative casing


sayhisam1:
Referencing https://github.com/Roblox/rodux/issues/48

It would be helpful to support both "disconnect" and "destroy" as alternative cases used in the maid class, since these conform to the camelCase naming convention used by popular libraries such as rodux.

Quenty:
Probably won't fix this one. This adds a lot of documentation and clarity burden, and makes refactoring harder. If you want to add a wrapper to automatically generate this, it should not be too much code.

PR #157, opened 2020-08-03
Add casing aliases for Maid class methods

ReturnedTrue:
First ever PR to a project, hope I did fine.

Fixes issue: #147

### DIFF
--- a/Modules/Client/Gui/GuiVisibleManager.lua
+++ b/Modules/Client/Gui/GuiVisibleManager.lua
@@ -80,6 +80,10 @@ function GuiVisibleManager:CreateShowHandle()
 
 	return {
 		Destroy = function()
+			if not self.Destroy then
+				return
+			end
+
 			if self._showHandles[key] then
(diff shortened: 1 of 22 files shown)


colbert2677:
If you're adding camelCase aliases for all Maid methods, should you not also add a camelCase variant for DoCleaning? I feel like it'd be more complete of a job that way.

ReturnedTrue:
Good point, added it.

sayhisam1:
This should also add casing aliases for objects added to the maid (e.g. maid should check if an object has a `destroy` method, in addition to the current check for `Destroy`)

Quenty:
Probably not going to merge this, sorry! I had variable casing in Nevermore for a while but it became a pain to maintain.

In the future, I'd like to switch syntax to `:lowerCamelCase()` but this will require a significant refactor. Now that we have versioning, you should be able to fork maid pretty easily. We may also be able to reversion packages if we refactor to :lowerCamelCase().

PR #163, opened 2020-09-06
Replace "Nevermore" with "NevermoreEngine"

Hoidberg:


Quenty:
The code require() all references .Nevermore, so if we make this change, it will break all module loading. D:

PR #165, opened 2020-09-19
Update RagdollUtils.lua

Hoidberg:


Quenty:
We probably don’t need this change, you can use the maid as a cookie to clean up the state by calling maid:Destroy()

PR #167, opened 2020-09-21
Make fadeTime optional

Hoidberg:


Quenty:
It's already optional. If you don't pass in a parameter it defaults to 0.2.

Quenty:
I'm closing this because it doesn't change code. It might help with readability tho, but we can be more explicit in this case.

PR #169, opened 2020-09-25
Allow for requiring numerical modules

mobyboyy:
I've made a minor fix that allows for requiring numerical modules, however, it's somewhat hacky, and may not work well with nevermore.


Quenty:
I wrote a nonhacky version of this to function properly!

PR #170, opened 2020-10-05
Update ClientTranslator.lua

Hoidberg:
```lua
assert(RunService:IsClient(), "ClientTranslator can only be required on client")
```
is not needed because the <a href="https://github.com/Quenty/NevermoreEngine/blob/version2/loader/ReplicatedStorage/Nevermore/init.lua">module loader</a> already prevents serverscripts from accessing client modules


devSparkle (approved):
Code and reasoning checks out. Keeping these kind of useless checks adds overhead (even if minor) and propagates bad practices.

Hoidberg (commented):
I don't know what to put here lol

Quenty (approved):
Nice

PR #175, opened 2020-11-16
Update RagdollUtils.lua

Hoidberg:


Hoidberg (commented):
Forgot to remove a whitespace-only line

PR #176, opened 2020-11-23
replace spawn with coroutine in MasterClock.lua

jakebball:
replaces spawn with coroutine.wrap for immediate start and future proofing.


T-R-I-X (approved):
Better optimization than spawn().

Quenty:
if we're using wait(5) then there's really no point here in using coroutine.wrap().

PR #180, opened 2020-12-30
Fixed Installer

howmanysmall:
Now it will install Nevermore properly, as well as using `RequestAsync` instead of `GetAsync` (faster at getting things for the most part).


howmanysmall:
luacheck seems to have failed on something i didn't change

Quenty:
Closing, installer was fixed recently.

PR #193, opened 2021-05-31
Add ./Modules/Shared/Physics/UnitUtils.lua

cassis163:
UnitUtils.lua makes converting numbers that have meters and or kilograms as units from the real world to Roblox and visa versa much easier. This can be useful when importing real world numbers like 997 kg/m^3 (which is roughly the density of sea water) to Roblox.

The correctness of UnitUtils can be tested. For instance, you can simulate the buoyancy of a part with a variable size in a medium. The part should have the exact same density as the medium. If the calculations are correct, the part would stay still in the medium.

One limitation of this class is that it only works well when 'workspace.Gravity' is equal to 35. This constant is predefined in the 'Realistic' setting of 'World Settings'. It will still work, but the scale of each meter would be unrealistic.

### DIFF
--- a/Modules/Shared/Physics/UnitUtils.lua
+++ b/Modules/Shared/Physics/UnitUtils.lua
@@ -1,0 +1,35 @@
+-- Library that is capable of converting real world to Roblox units and visa versa.
+-- @module UnitUtils
+
+local UnitUtils = {}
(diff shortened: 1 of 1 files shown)


Quenty:
See MetricUtils.lua, and try adding to that!

PR #222, opened 2021-11-09
Create wally.toml

brandon-wilcox-supersocial:
Create a Wally.toml to make this a valid Wally package


Quenty:
You'll need to rename the title to ci: Create wally.toml for Signal package or something

PR #233, opened 2021-12-10
use common static values for Vector2/Vector3/CFrame

Baileyeatspizza:
in a recent [release note](https://developer.roblox.com/en-us/resources/release-note/Release-Notes-for-502) 
Screenshot 2021-12-10 at 20 21 32

this pull request basically just replaces any instances where those static methods can be used over constructing a new datatype

here's a benchmark of the performance gains of using Vector3.zero instead of Vector3.new(...) 
image
benchmarked on an Apple M1


Quenty:
Looks great. We'll need to update luacheck to make this happen.

Quenty:
And probably update selene too.

Baileyeatspizza:
seems like the new methods aren't in the lint check so ignore that check

PR #246, opened 2022-01-15
fix: Disabled CanTouch and CanQuery

Yuuwa0519:
Parts created in draw module shouldn't interfere with anything, therefore I thought CanTouch and CanQuery should also be false.


Yuuwa0519:
Sorry for making lots of naming changes. I have never created pull request and wasn't too sure of what kind of change this is, so I changed it around multiple times.

Quenty:
Thank you for your contribution!

PR #249, opened 2022-01-27
feat: Added basic Attribute support + documentation to Blend

nightcycle:


src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -548,6 +575,48 @@ function Blend.Children(parent, value)
...
+]=]
+function Blend.Attributes(inst, value)
+	assert(typeof(inst) == "Instance", "Bad instance")
+	return Blend.ComputedPairs(value, function(attributeName, attributeValue, innerMaid)
+		local function getValue(v)
+			if type(v) == "table" then
+				if ValueObject.isValueObject(v) then
+					return getValue(v.Value)
+				elseif Brio.isBrio(v) then
+					return getValue(v:GetValue())
```
Quenty, line 607: nit: Should probably keep lifetime of attribute application to the brio lifetime. I think you can do 

```lua
innerMaid[attributeName] = v:GetDiedSignal():Connect(function()
    inst:SetAttribute(attributeName, attributeValue)
end)
```

You'll also need to set
innerMaid[attributeName] = nil

Before you do each item.


This should only occur if we emit a brio with another brio in it, so we could also remove the `if Brio.isBrio` condition entirely. Up to you.

src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -548,6 +575,48 @@ function Blend.Children(parent, value)
...
+	Cleanup:
+	* Attributes will be cleaned up on unsubscribe
+
+	@param parent Instance
+	@param value any
+	@return MaidTask
+]=]
+function Blend.Attributes(inst, value)
+	assert(typeof(inst) == "Instance", "Bad instance")
+	return Blend.ComputedPairs(value, function(attributeName, attributeValue, innerMaid)
```
Quenty, line 601: Might be good to error if attributeName is not a string.

src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -548,6 +575,48 @@ function Blend.Children(parent, value)
...
+--[=[
+	Mounts attributes to the parent and returns an object which will cleanup and delete
+	all attributes when removed.
+
+	```lua
+	local attribute = Blend.State("abc")
+	Blend.New "ScreenGui" {
+		Parent = game.Players.LocalPlayer.PlayerGui;
+		[Blend.Attributes] = {
+			Attribute1 = attribute;
```
Quenty, line 587: Unfortunately, ComputedPairs does not let us take in an [Observable] = [Observable] yet, which would make sense in certain situations.

I think this can be a future problem.

src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -548,6 +575,48 @@ function Blend.Children(parent, value)
...
+			Attribute2 = 123;
+		};
+	};
+	```
+	Cleanup:
+	* Attributes will be cleaned up on unsubscribe
```
Quenty, line 593: nit: would reword this to `Attributes will be set to nil on unsubscribe`.

src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -1,3 +1,4 @@
+---@diagnostic disable: redefined-local
```
Quenty, line 1: nit: We should not disable this redefined-local lint, and instead we should actually not redefine a local value.

src/blend/package.json
```diff
@@ -21,7 +21,8 @@
   },
   "license": "MIT",
   "contributors": [
-    "Quenty"
+    "Quenty",
+    "Nightcycle"
```
Quenty, line 25: Make sure you have an NPM account :p

Quenty:
If you want to also update the `plugin/ui-converter-plugin` to support this, that would be neat, but not necessary.

Quenty (changes requested):
Looks great overall, left a ton of nit-picks. If you don't want to address them, let me know, and I can do it. :p

src/blend/src/Shared/Blend/Blend.lua
nightcycle: I'm still getting my sea legs with brio, so I left the if Brio.isBrio for now, but you're welcome to change it. I believe I've ported the rest of the code though!

nightcycle:
> If you want to also update the `plugin/ui-converter-plugin` to support this, that would be neat, but not necessary.

I agree that's a good idea however after looking over the code I admit it unfortunately might be above my current ability level.

src/blend/src/Shared/Blend/Blend.lua
nightcycle: In that case I've added a bit of extra documentation here to better convey that + an optional workaround for now.

src/blend/src/Shared/Blend/Blend.lua
Quenty: Sounds good.

Quenty (approved):

PR #259, opened 2022-04-25
feat: Add thread-support for maids

optimisticside:
Maids should call `task.cancel` on jobs it's given that are threads.


Quenty (approved):

Quenty:
need to also update MaidTaskUtils.lua

PR #262, opened 2022-06-12
feat: Add RxCharacterUtils

OttoHatt:
I didn't like doing `Humanoid.Health > 0`  every frame, so I made some small utilities for it.
Sorry for not doing a changelog! Also will the package dependencies will need updating too??

### DIFF
--- a/src/characterutils/src/Shared/RxCharacterUtils.lua
+++ b/src/characterutils/src/Shared/RxCharacterUtils.lua
@@ -1,0 +1,82 @@
+--[=[
+	Utilities for observing characters (and humanoids).
+	@class RxCharacterUtils
+]=]
+
+local require = require(script.Parent.loader).load(script)
+
+local Brio = require("Brio")
+local Maid = require("Maid")
+local Observable = require("Observable")
+local RxBrioUtils = require("RxBrioUtils")
+local RxInstanceUtils = require("RxInstanceUtils")
+
+local RxCharacterUtils = {}
+
+--[=[
+	Observe a player's last character.
+	@param player Player
+	@return Observable<Brio<Model>>
+]=]
+function RxCharacterUtils.observeLastCharacterBrio(player: Player)
+	return RxInstanceUtils.observePropertyBrio(player, "Character"):Pipe({
+		RxBrioUtils.where(function(character)
+			return character ~= nil
+		end),
+	})
+end
+
+--[=[
+	Observe a player's last humanoid. Note that it may not be alive!
+	@param player Player
+	@return Observable<Brio<Humanoid>>
+]=]
+function RxCharacterUtils.observeLastHumanoidBrio(player: Player)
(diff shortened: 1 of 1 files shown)


src/characterutils/src/Shared/RxCharacterUtils.lua
```diff
@@ -0,0 +1,82 @@
...
+		RxBrioUtils.switchMapBrio(function(character)
+			return RxInstanceUtils.observeLastNamedChildBrio(character, "Humanoid", "Humanoid")
+		end),
+	})
+end
+
+--[=[
+	Returns an observable that emites a single brio with the value of the given humanoid.
+	When the humanoid dies, the brio is killed and the subscription completes.
+	If the humanoid is dead on subscription, the subscription immediatelty completes with nothing emitted.
```
Quenty, line 45: nit: Spelling of immediately

src/characterutils/src/Shared/RxCharacterUtils.lua
```diff
@@ -0,0 +1,82 @@
...
+			sub:Complete()
+		end
+
+		if humanoid.Health > 0 then
+			local maid = Maid.new()
+
+			maid._brio = Brio.new(humanoid)
+			sub:Fire(maid._brio)
+
+			maid:GiveTask(humanoid.Died:Connect(onDeath))
```
Quenty, line 62: side note: .Died can fire multiple times, but it's ok because we disconnect immediately.

src/characterutils/src/Shared/RxCharacterUtils.lua
```diff
@@ -0,0 +1,82 @@
...
+--[=[
+	Observe a player's last character.
+	@param player Player
+	@return Observable<Brio<Model>>
+]=]
+function RxCharacterUtils.observeLastCharacterBrio(player: Player)
+	return RxInstanceUtils.observePropertyBrio(player, "Character"):Pipe({
+		RxBrioUtils.where(function(character)
+			return character ~= nil
+		end),
```
Quenty, line 25: nit: You can put this check like this to save some perf

```lua
return RxInstanceUtils.observePropertyBrio(player, "Character", function(character)
	return character ~= nil
end)
```

Quenty (approved):

Quenty:
Looks good overall, I'm ok with merging as is, but left a few comments just in case you wanted to address them.

Quenty:
No changelog necessary, it is auto-generated from commit. Same with dependency updates. This looks good.

OttoHatt:
Added your improvements!
This looks like it'll have some overlap with [HumanoidTracker](https://github.com/Quenty/NevermoreEngine/blob/9f7eaea7543c33c89d2e32c38491b13f9271f4f7/src/humanoidtracker/src/Shared/HumanoidTracker.lua), but I don't think it's worth moving it to Rx?

OttoHatt:
All good now, let me know if there’s anything else to change before merging.

PR #264, opened 2022-06-18
Optimise + document Table library

OttoHatt:
Updated some of the table util functions to use the newer `table.` methods. Hopefully they'll be faster!(?). Also changed some doc phrasing.

I've done my best to test, but I know changes to a module like this can incur a lot of hard-to-find and obscure bugs (likely to do with mix/matching dictionary-style and array-style tables). Consider testing thoroughly with older nevermore projects, or simply not merging????

### DIFF
--- a/src/table/src/Shared/Table.lua
+++ b/src/table/src/Shared/Table.lua
@@ -295,6 +295,7 @@ local READ_ONLY_METATABLE = {
 	Sets a metatable on a table such that it errors when
 	indexing a nil value
 
+	@ignore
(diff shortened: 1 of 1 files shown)


OttoHatt:
It may be a good idea to rename the `.readonly` methods; they provide the nice functionality in erroring on a nil index, but I think it's easy to assume it does the same as `table.freeze` (which errors on write but not on nil indexing).

Quenty:
Closed in favor of #321

PR #267, opened 2022-06-21
Update SoftShutdownServiceClient.lua to fix ScreenGui not disabled.

antcar0929:
👋Hi! This is my first time contributing to an Open Source project, I don't really know how to contribute on GitHub, so I will make mistakes. 

-----

This PR is to fix the problem of SoftShutdownService not disabling UIs that are created after `_hideCoreGuiUI` is called. Also, it fixes that the soft shutdown UI not being on top, since `1E10` on a ScreenGui DisplayOrder will be solved as `-2147483648` (The value is too big, I don't know if `1E10` is meant to be there).

-----

unknown
^The ScreenGui is created after `_hideCoreGuiUI` is called.

-----

I've tested my own code before opening a pull request.

### DIFF
--- a/src/actionmanager/CHANGELOG.md
+++ b/src/actionmanager/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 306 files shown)


Quenty:
Thank you! I fixed this in another commit before seeing this! 

If you edit this and JUST do the 1e9 that would be neat.

PR #271, opened 2022-07-18
fix: Fix broken initial value of Blend.AccelTween

OttoHatt:
Blend was passing the initial value into the constructor of AccelTween, which actually sets its acceleration. This has probably caused lots of spooky behaviour! `Blend.AccelTween` will now use the correct acceleration, and won't always start at zero.


OttoHatt:
oops, accidentally used a junk email for the commit, fixed now

Quenty (approved):

PR #272, opened 2022-07-24
Datastore UTF-8 support

OttoHatt:
Nevermore's datastore API is awesome, but a limitation I've found is that it requires you to pass a `ValueBase` into `:StoreOnValueChange(...)`. The problem with this is that `StringValue`s can't hold UTF-8 data, which rules out using libraries like [MessagePack](https://github.com/cipharius/msgpack-luau).

As a quick fix I allowed passing a ValueObject into DataStoreStage, but I don't think it's a good solution as then you could pass in all manner of unserializable tables or instances. Maybe this method should check that the value is always a string before saving? Unsure.

### DIFF
--- a/src/datastore/src/Server/Modules/DataStoreStage.lua
+++ b/src/datastore/src/Server/Modules/DataStoreStage.lua
@@ -17,6 +17,7 @@ local Promise = require("Promise")
 local PromiseUtils = require("PromiseUtils")
 local Signal = require("Signal")
 local Table = require("Table")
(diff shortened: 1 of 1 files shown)


OttoHatt:
Example of a string that won't save properly: 

OeUC=ZeLV%&AC=ZekBRC4\L7+%hr<i2@UV \T4[,W~

Quenty:
Fixed by #276

PR #275, opened 2022-07-30
fix: Disable CanQuery/CanTouch on debug line part

OttoHatt:
Fixes a typo from copy + pasting


Quenty:
wow, github UI always adds a new line at the end of the file. very unfortunate.

PR #282, opened 2022-08-15
feat: Add TintController package for colouring groups of instances

OttoHatt:
Provides a class that can be used to colour groups of tagged instances at once. Useful as it supports most useful instances with a Color3 property, like `BasePart` `Texture` `ImageLabel` etc.

I've never written a package before! Please give feedback and changes.

### DIFF
--- a/src/tintcontroller/Client/TintBindersClient.lua
+++ b/src/tintcontroller/Client/TintBindersClient.lua
@@ -1,0 +1,37 @@
+--[=[
+	Initializes the tinting system on the client.
+
+	```lua
+	local TintBindersClient = require("TintBindersClient")
+	local TintControllerUtils = require("TintControllerUtils")
+	local ServiceBag = require("ServiceBag")
+
+	local serviceBag = ServiceBag.new()
+	serviceBag:GetService(TintBindersClient)
+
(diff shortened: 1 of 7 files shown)


OttoHatt:
Changed the behaviour to use a blending mode on the model, instead of outright setting its colour. Currently it uses a soft light blending mode between the tint colour and the part's original colour.

However I would like the behaviour to be such that when alpha = 0, the instance is its original colour. And when alpha = 1, every descendant is set to exactly that of the tint colour. A value inbetween would make the colours transition. I don't know of a blending mode that does this though!

Quenty:
I'll take a look in a bit, but it looks like we're missing package-specific instances?

OttoHatt:
Oops! I forgot about those.

Had a look over this again. This package automatically tints an instance's colourable descendants using a combination of binders and `.AncestryChanged`. This method isn't perfect and leaves some strange behaviour (like what if you parent/unparent, recolour after parenting, stream in/out, etc).

I think to have a performant 'tinting' feature (i.e. without several connections on every descendant) you'd need to build it around your use case, incurring lots of compromises and undefined behaviours in the process. I don't think it'll ever be general purpose enough!

PR #284, opened 2022-08-17
feat: Free DataStore key after its writer GCs to allow future writers

OttoHatt:
Modifies `DataStoreStage.StoreOnValueChange` to also free a key after removing a datastore writer. This still allows only one writer at a time, but prevents a bug caused by attaching a new writer when a player re-joins the game.


OttoHatt:
Good to merge?

PR #291, opened 2022-10-09
Fix SlerpQuaternions

EthanCurtiss:
It appears `SlerpQuaternions` was intended to follow either the first or fourth of the following equivalent equations, but the order of the arguments to one of the Qmul calls was switched.
image
[https://en.wikipedia.org/wiki/Slerp#Quaternion_Slerp](https://en.wikipedia.org/wiki/Slerp#Quaternion_Slerp)
I chose to convert it to the fourth equation.


Quenty:
Ah, release pipeline is broken right now for pull_request scenarios. Need to patch it! Will merge soon.

PR #294, opened 2022-10-10
fix: Fix QFrame.isNaN

EthanCurtiss:
`QFrame.isNaN` returns the opposite boolean that it should. This fixes that.


Quenty:
oh my gosh, you're right! That's incredibly awkward.

Quenty:
Ah, release pipeline is broken right now for pull_request scenarios. Need to patch it! Will merge soon.

Quenty:
No idea why it shipped all of these instead of just 1 package. Will need to investigate later.

PR #321, opened 2023-01-14
feat!: Favoring Lua table methods

OttoHatt:
Fixes all occurrences of Selene's 'manual_table_clone' lint. Also has slightly better docs.

**BREAKING CHANGE**: This breaks table.getIndex in the metatable scenario of comparison using `==`. Generally this contract should not actually cause any issues but it's important that we identify this as an issue.

This will cause a wide variety of package releases in major version.

### DIFF
--- a/src/table/src/Shared/Table.lua
+++ b/src/table/src/Shared/Table.lua
@@ -80,16 +80,16 @@ end
 --[=[
 	Shallow merges two lists without modifying either.
 
-	@param orig table -- Original table
-	@param new table -- Result
+	@param a table -- Source table
+	@param b table -- Source table
 	@return table
 ]=]
-function Table.mergeLists(orig, new)
+function Table.mergeLists(a, b)
 	local _table = {}
-	for _, val in pairs(orig) do
+	for _, val in pairs(a) do
 		table.insert(_table, val)
 	end
-	for _, val in pairs(new) do
+	for _, val in pairs(b) do
 		table.insert(_table, val)
 	end
 	return _table
--- a/src/table/src/Shared/Table.lua
+++ b/src/table/src/Shared/Table.lua
@@ -197,16 +197,7 @@ end
 	@return The index of the value, if found
 	@return nil -- if not found
 ]=]
-function Table.getIndex(haystack, needle)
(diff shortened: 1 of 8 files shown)


src/table/src/Shared/Table.lua
```diff
@@ -200,16 +197,7 @@ end
...
-	assert(needle ~= nil, "Needle cannot be nil")
-
-	for index, item in pairs(haystack) do
-		if needle == item then
-			return index
-		end
-	end
-	return nil
-end
+Table.getIndex = table.find
```
Quenty, line 200: This may actually be a breaking change, does table.find respect metatableS? probably still fine. may need to increment version of table library though.

Quenty (approved):

src/table/src/Shared/Table.lua
Quenty: I suspect we should major-version this package.

OttoHatt: Good catch. Looks like it does, here's a snippet.
```lua
local function getIndex(haystack, needle)
	for i, v in pairs(haystack) do
		if needle == v then
			return i
		end
	end
end

local mt = {}
mt.__eq = function(a, b)
	return true
end

local a = setmetatable({}, mt)
local b = setmetatable({}, mt)

local testArr = {a, b}

-- 1, 1.
print(getIndex(testArr, b), table.find(testArr, b))
-- 2, 2.
mt.__eq = nil
print(getIndex(testArr, b), table.find(testArr, b))
```

Quenty:
Can we update selene to also check for this in the future?

Quenty:
Because we use table in so many places, this breaking change will propagate throughout the other packages. This is probably fine, but I just wanted to note that is why there is a slight delay here.

OttoHatt:
Did you see the snippet I posted? I found that `table.find` respects metatables.

OttoHatt:
https://github.com/Roblox/luau/blob/master/rfcs/behavior-eq-metamethod.md

OttoHatt:
This is mostly resolved in the latest commits

PR #324, opened 2023-01-17
fix: 'Promise.isPromise' checks metatable over reading ClassName

OttoHatt:
I've been wanting to return `Table.readOnly` tables from inside promises. However I've run into an issue: every value returned from a Promise is checked against `Promise.isPromise` for chaining, which reads `ClassName` directly and causes the `readonly` metatable to throw. My solution is to use `getmetatable` instead, like newer Nevermore classes.

### DIFF
--- a/src/promise/src/Shared/Promise.lua
+++ b/src/promise/src/Shared/Promise.lua
@@ -22,7 +22,7 @@ Promise.__index = Promise
 	@return boolean
 ]=]
 function Promise.isPromise(value)
-	return type(value) == "table" and value.ClassName == "Promise"
+	return type(value) == "table" and getmetatable(value) == Promise
 end
 
 --[=[


Quenty:
I'm good for this.

I think we should generalize the isA() calls to be duck typing instead of this classname checks.

Quenty:
Oh wait, this can technically break stuff because of the way Rx works with multiple copies of promises. I'm still ok with this change, but duck typing would be better. Probably need a `Promiseinterface` which works with eryn's promise impl. Could standardize a spec within Nevermore.

OttoHatt:
Switching to `getmetatable(x).ClassName == "Promise"` may be a good enough workaround for now? Metatables typically aren't protected.

Quenty:
I think so! We're probably going to need to dummy-type the Promise for compatibility with Eryn's promise soon anyway.

OttoHatt:
Looks like 14874e8f4b0789e203bd60f418d70510fac950e9 added ducktyping. Worth looking at again? Not sure of the performance implications over just checking the metatable.

PR #382, opened 2023-07-02
Blend.Single changes

OttoHatt:
* Allow returning `nil` from the observable. We know that it will never evaluate to anything interesting, so there's no point creating a brio (which are not free). (Should also probably skip creating a brio if the passed observable is complete and evaluates to nil, i.e. == `Rx.EMPTY`?).
* Propagate state of passed observable. This has just become a slightly more efficient version of a switchMap + toBrio.
* Returning from a subscription does nothing.

### DIFF
--- a/src/blend/src/Shared/Blend/Blend.lua
+++ b/src/blend/src/Shared/Blend/Blend.lua
@@ -787,14 +787,13 @@ function Blend.Single(observable)
 				local copy = BrioUtils.clone(result)
 				maid._current = copy
 				sub:Fire(copy)
-				return copy
+			elseif result then
+				local current = Brio.new(result)
+				maid._current = current
+				sub:Fire(current)
+			else
+				maid._current = nil
 			end
-
-			local current = Brio.new(result)
-			maid._current = current
-			sub:Fire(current)
-
-			return current
 		end))
 
 		return maid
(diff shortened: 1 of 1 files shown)


src/blend/src/Shared/Blend/Blend.lua
```diff
@@ -787,14 +787,17 @@ function Blend.Single(observable)
...
+			else
+				maid._current = nil
 			end
-
-			local current = Brio.new(result)
-			maid._current = current
-			sub:Fire(current)
-
-			return current
+		end, function(...)
```
Quenty, line 797: we can pass sub:GetFailComplete() here. Note I think it's fine to complete this here, but a failure could cause other observables in the blend tree to also fail. Most of Nevermore doesn't really leverage the failure state that much.

Quenty: If you think it's safe to merge this, I'm down for it.

Quenty:
Side note: Rx.EMPTY never completes I believe.

src/blend/src/Shared/Blend/Blend.lua
OttoHatt: I wanted the completion, I didn't realise this would could pass failures. I was only vaguely aware subscriptions had this - I know for certain I'm not handling it!

Though it's a breaking change, I would want to propagate this. I think hiding errors is bad. But silently failing is also bad, because I'm not handling this case and wouldn't know why my subscription stopped. I think subscriptions should print a warning (similar to promises) for an unhandled failure state, and then this should be merged?

OttoHatt:
If propagating failure is contentious, is it possible to merge just the first of these commits for now?

PR #386, opened 2023-07-11
fix: Pass 'memberName' in Remoting::Bind on the client

OttoHatt:


OttoHatt (commented):
I assume invoking the client is fine with this class? Everything should be wrapped in a promise...

PR #392, opened 2023-07-15
fix: Return results from non-promisified Remoting callbacks

OttoHatt:


OttoHatt (commented):
Codespaces makes me leave a comment here

OttoHatt:
It won't let me remove the newline at the end of the file :/

PR #423, opened 2023-10-14
fix: Fix function implementations/names in ObservableList

InvisibleWater:
This was bugging me for a while.

### DIFF
--- a/src/observablecollection/src/Shared/ObservableMapList.lua
+++ b/src/observablecollection/src/Shared/ObservableMapList.lua
@@ -204,7 +204,7 @@ end
 	@param key TKey
 	@return { TValue }
 ]=]
-function ObservableMapList:GetListForKey(key)
+function ObservableMapList:GetListFromKey(key)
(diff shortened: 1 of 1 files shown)


Quenty:
Oh oops.

src/observablecollection/src/Shared/ObservableMapList.lua
```diff
@@ -204,7 +204,7 @@ end
 	@param key TKey
 	@return { TValue }
 ]=]
-function ObservableMapList:GetListForKey(key)
```
Quenty, line 207: ok wait this is a breaking change.

Quenty: can we keep this one the same so it doesn't break old apis?

Quenty: oh wait, that's not a breaking change.

Quenty (approved):

PR #426, opened 2023-10-24
fix: Fix value removal function in ObservableMapList

InvisibleWater:
Just noticed that the methods called in here don't actually exist in ObservableList at the moment. I probably would have included this in the pull request I made earlier if I had caught this then.


Quenty:
Thanks! I think I accidentally patched this same thing.

PR #428, opened 2023-10-28
fix(rx): Rx.timer() bad coroutine close

MrAsynchronous:
* fixes issue in which Rx.timer() would fail on cleanup due to a bad coroutine.close() call


src/rx/src/Shared/Rx.lua
```diff
@@ -1883,4 +1883,4 @@ function Rx.throttle(durationSelector)
 	end
 end
 
-return Rx
\ No newline at end of file
+return Rx
```
Quenty, line 1886: Pls no

Quenty (changes requested):

MrAsynchronous (commented):
chore: removed whitespace

PR #430, opened 2023-11-22
fix: Don't update ButtonHighlightModel held count post-destruction

OttoHatt:
Fixes holding down a button as it's destroyed throwing `ButtonHighlightModel:399: attempt to perform arithmetic (sub) on nil and number`. Destruction order is temperamental, sometimes the `_numFingerDown` ValueObject is destroyed before the callback runs.


Quenty:
I think I patched this separately. This is good now, right?

OttoHatt:
Yes! Same fix.

PR #443, opened 2024-02-05
feat[Viewport]: Added ability to disable ViewportControls

MrAsynchronous:
* Adds API to allow consumers to disable / enable both `ViewportControl` instance, and `Viewport` instances

### DIFF
--- a/src/viewport/src/Client/Viewport.lua
+++ b/src/viewport/src/Client/Viewport.lua
@@ -52,6 +52,7 @@ function Viewport.new()
 	self._transparency = self._maid:Add(ValueObject.new(0, "number"))
 	self._absoluteSize = self._maid:Add(ValueObject.new(Vector2.zero, "Vector2"))
 	self._fieldOfView = self._maid:Add(ValueObject.new(20, "number"))
+	self._controlsEnabled = self._maid:Add(ValueObject.new(true, "boolean"))
 
 	self._rotationYawSpring = self._maid:Add(SpringObject.new(math.rad(90 + 90 - 30)))
 	self._rotationYawSpring.Speed = 30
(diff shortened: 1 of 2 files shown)


src/viewport/src/Client/ViewportControls.lua
```diff
@@ -25,18 +26,31 @@ function ViewportControls.new(viewport, viewportModel)
 	local self = setmetatable(BaseObject.new(viewport), ViewportControls)
 
 	self._viewportModel = assert(viewportModel, "No rotationYaw")
+	self._enabled = self._maid:Add(ValueObject.new())
```
Quenty, line 29: Unfortunately we either need this to be a breaking change and disable by default, OR swap this to true. Either way we should include the type annotation.

```suggestion
	self._enabled = self._maid:Add(ValueObject.new(true, "boolean"))
```

src/viewport/src/Client/ViewportControls.lua
```diff
@@ -25,18 +26,31 @@ function ViewportControls.new(viewport, viewportModel)
...
 
+--[=[
+	Sets the enabled state of the controls
+
+	@param enabled boolean
+]=]
+function ViewportControls:SetEnabled(enabled)
+	assert(type(enabled) == "boolean", "Bad enabled")
+
+	self._enabled.Value = enabled
```
Quenty, line 51: you could do

```lua
return self._enabled:Mount(enabled)
```

and remove the assert if you wanted to just mount an observable directly here. then in high level call you can do:

```lua
self._controls:SetEnabled(self._isEnabled:Observe())
```

no change necessary. thought you would appreciate that tho.

src/viewport/src/Client/Viewport.lua
```diff
@@ -253,7 +265,16 @@ function Viewport:Render(props)
 			end);
 		[Blend.OnChange "AbsoluteSize"] = self._absoluteSize;
 		[Blend.Attached(function(viewport)
-			return ViewportControls.new(viewport, self)
+			local controlsMaid = Maid.new()
+
+			-- create viewport controls and obey enabled state
+			local viewportControls = ViewportControls.new(viewport, self)
```
Quenty, line 271: might technically be nicer to construct only when necessary, but this code is fine tbh.

Quenty (approved):

PR #464, opened 2024-04-29
Add FastSignal as an alternative/replacement to GoodSignal

### LINKED ISSUE #463: Implement Deferred/Adapted Signal using Lua
**What package is this for?**
Signal

**Describe your problem**
Currently, Signal by defaults uses GoodSignal which has some issues but is nice, but it also has the option of the older standard that uses BindableEvents. BindableEvents allow for Deferred (Adaptive) behavior which is important, GoodSignal doesn't. At the same time however, BindableEvents suck. They cause memory leaks which may not be detected by the user.

**Describe the solution you'd like**

We should abandon the BindableEvent approach. It is slower, leakier, just plain worse. We now have the power of Lua Signals, have had so for a long time. But it is also important to implement Deferred behavior.

We have task.defer, and it is also possible to infer what mode the game is in with some simple code.
I did this in [my Signal library](https://github.com/RBLXUtils/FastSignal/tree/main/src/ReplicatedStorage/FastSignal) which is useful to implementing this. You can use it if you like (I don't maintain it, running Linux can't anymore, it doesn't have the same typing as sleitnick's fork for example) or fork/use the ideas/code from there. (I think so anyway? I haven't looked at licenses for a while but I can probably change it for this if it is an issue)

Thank you for your time.

lucasmz-dev:
FastSignal when referred here is not the alternative provided to GoodSignal by their own post. This is a library I made; It isn't actually as speed focused as GoodSignal, despite the name, I've been thinking of changing it to BetterSignal but that just sounds problematic... Anyhow..

This implements a proposed solution at solving #463.

This pull request has as its objective:
* Add FastSignal
* Default to it

This could actually replace both, the Signal.lua implementation, and GoodSignal, but at this time I'm only adding it as an alternative, I am open to changing this on how it is decided to implement this.

Need review:
* If my code is correct, I do not necessarily use NevermoreEngine itself, and I'm not sure if my modifications make sure inside the engine, please review.

Fixes #463

### DIFF
--- a/src/signal/src/Shared/FastSignal/Deferred.lua
+++ b/src/signal/src/Shared/FastSignal/Deferred.lua
@@ -1,0 +1,342 @@
+--[=[
+	A class which holds data and methods for ScriptSignals.
+
+	@class ScriptSignal
+]=]
+local ScriptSignal = {}
+ScriptSignal.__index = ScriptSignal
+
+--[=[
+	A class which holds data and methods for ScriptConnections.
+
+	@class ScriptConnection
+]=]
+local ScriptConnection = {}
+ScriptConnection.__index = ScriptConnection
+
+--[=[
+	A boolean which determines if a ScriptConnection is active or not.
+
+	@prop Connected boolean
+	@within ScriptConnection
+
+	@readonly
+]=]
+
+export type Class = typeof( setmetatable({
+	_active = true,
+	_head = nil :: ScriptConnectionNode?
+}, ScriptSignal) )
+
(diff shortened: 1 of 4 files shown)


lucasmz-dev:
I am not really able to test things here out, I am not really in Roblox development and can't be anymore properly due to Roblox's decision to not support Linux :(, just noticed this issue and wanted to note to it. Therefore, the changes made here are minor so that there aren't any issues I can't debug and such.

PR #465, opened 2024-05-01
feat: Add :GetTimeRemaining() to DebounceTimer

InvisibleWater:


src/debounce/src/Shared/DebounceTimer.lua
```diff
@@ -43,6 +43,18 @@ function DebounceTimer:IsRunning()
...
+--[=[
+	Returns the amount of time remaining in the timer.
+	@return number
+]=]
+function DebounceTimer:GetTimeRemaining()
+	if not self:IsRunning() then
+		return 0
+	end
+
+	return math.max((self._startTime + self._length) - tick(), 0)
```
Quenty, line 55: also valid to clamp with length, I think, but this is good

PR #470, opened 2024-05-13
fix: Using integer keys in w/ Observe methods in ObservableMap

OttoHatt:
The maid is given some cleanup w/ `:GiveTask`, so they get given keys `1` and `2` internally. The value brios are also put into this maid, so if you did `map:Set(1, ...)` you'd break everything observing the container.


Quenty:
Let's make a new maid specifically for the keys instead of trying to fix the :GiveTask() call. That way future changes won't break anything. 

We should also maybe add a method to Maid that allows you to get/set a task that could conflict with a name. For example. `GiveTask` as a key could be really bad. 

We should consider adding maid:Set() or something for safety.

Quenty (approved):

PR #509, opened 2024-10-13
ObservableList.fromObservableBrio and ObservableSet.fromObservableBrio.

InvisibleWater:
I noticed that something similar to RxStateStackUtils.createStateStack could be done with ObservableLists, and that eventually led to this.


InvisibleWater:
Now that I think about it, there's definitely a better way to implement this.

PR #511, opened 2024-10-19
ValueObject - Fix of typo in initial type check

InvisibleWater:


Quenty:
feel like a unit test should have caught this

PR #512, opened 2024-10-20
feat: Rx.skipUntil, Rx.skipWhile, and Rx.takeWhile

InvisibleWater:


src/rx/src/Shared/Rx.lua
```diff
@@ -1947,4 +1947,99 @@ function Rx.throttle(durationSelector)
...
+--[=[
+	Skips over values emitted by the source observable until a passed-in notifier observable emits a value.
+
+	https://rxjs.dev/api/operators/skipUntil
+
+	@param notifier Observable
+	@return (source: Observable) -> Observable
+]=]
+function Rx.skipUntil(notifier)
+	assert(Observable.isObservable(notifier))
```
Quenty, line 1959:
```suggestion
	assert(Observable.isObservable(source), "Bad observable")
```

src/rx/src/Shared/Rx.lua
```diff
@@ -1947,4 +1947,99 @@ function Rx.throttle(durationSelector)
...
+	@return (source: Observable) -> Observable
+]=]
+function Rx.takeWhile(predicate)
+	assert(type(predicate) == "function", "Bad predicate")
+
+	return function(source)
+		assert(Observable.isObservable(source), "Bad observable")
+
+		return Observable.new(function(sub)
+			local index = 1
```
Quenty, line 2031: shouldn't this start at 0?

Quenty (commented):
looks good overall!

src/rx/src/Shared/Rx.lua
InvisibleWater: Probably. I might have just thrown that in there randomly when I noticed that it first increments after the predicate is called.

I ended up changing it so that the index starts at 0, then is incremented before checking to see if the predicate is true.

PR #555, opened 2025-04-25
fix: Support Content type in Blend converter plugin

unrooot:
Fixes a memory leak with selected instances, and adds support for the new `Content` type. On ImageLabels, the `ImageContent` property is added in addition to `Image`, so I decided to disable it for now since it's not required.

### DIFF
--- a/plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
+++ b/plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
@@ -91,9 +91,9 @@ function UIConverterUtils.toLuaPropertyString(value: any, debugHint: string): st
 	if valueType == "string" then
 		local multiline = UIConverterUtils.toMultiLineEscape(value)
 		if multiline then
-			return multiline
+			return `"{multiline}"`
 		else
-			return string.format("%q", value)
+			return `"{value}"`
 		end
 	elseif valueType == "number" then
 		return roundNumber(value)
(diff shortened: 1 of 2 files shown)


Quenty (approved):

plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
```diff
@@ -91,9 +91,9 @@ function UIConverterUtils.toLuaPropertyString(value: any, debugHint: string): st
 	if valueType == "string" then
 		local multiline = UIConverterUtils.toMultiLineEscape(value)
 		if multiline then
-			return multiline
+			return `"{multiline}"`
 		else
-			return string.format("%q", value)
+			return `"{value}"`
```
Quenty, line 96: I think you still need %q here to escape properly

plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
```diff
@@ -269,7 +276,7 @@ end
 function UIConverterUtils.convertPropertiesToTable(properties, refLookupMap)
 	local data = {}
 	for key, value in properties do
-		if key ~= "Parent" then
+		if key ~= "Parent" and key ~= "ImageContent" then
```
Quenty, line 279: May want to handle the ImageContent as typeof(blah) == “content”) instead

Quenty (approved):

plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
unrooot: thx, I was wondering about this - fixed in [b1ef99c](https://github.com/Quenty/NevermoreEngine/pull/555/commits/b1ef99cd7bf481ffafc8c52c2277d04bad2299a1)

plugins/ui-converter-plugin/src/modules/Server/UIConverterUtils.lua
unrooot: good catch, thx - fixed in [b1ef99c](https://github.com/Quenty/NevermoreEngine/pull/555/commits/b1ef99cd7bf481ffafc8c52c2277d04bad2299a1)

PR #567, opened 2025-07-08
fix: Fix UniqueId error in Blend converter plugin

unrooot:
A recent Roblox update changed the write security of UniqueId to RobloxSecurity, which broke the plugin. These fixes should properly check for it and fix the error.


Quenty:
:rocket: PR was released in [`@quenty/roblox-api-dump@8.12.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/roblox-api-dump%408.12.1) :rocket:

PR #574, opened 2025-08-02
Fix: BoundParentTracker does not track parent being binded

duckymomo360:
Fix for probably unintended behavior.
[BoundAncestorTracker](https://github.com/Quenty/NevermoreEngine/blob/main/src/binder/src/Shared/Trackers/BoundAncestorTracker.lua#L45) tracks ancestors being binded, but BoundParentTracker does not track parents being binded.

### DIFF
--- a/src/binder/src/Shared/Trackers/BoundParentTracker.lua
+++ b/src/binder/src/Shared/Trackers/BoundParentTracker.lua
@@ -29,6 +29,12 @@ function BoundParentTracker.new(binder, child)
 		end
 	end))
 
+	self._maid:GiveTask(self._binder:GetClassAddedSignal():Connect(function(_, instance)
+		if self._child.Parent == instance then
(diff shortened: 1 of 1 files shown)


src/binder/src/Shared/Trackers/BoundParentTracker.lua
```diff
@@ -29,6 +29,12 @@ function BoundParentTracker.new(binder, child)
 		end
 	end))
 
+	self._maid:GiveTask(self._binder:GetClassAddedSignal():Connect(function(_, instance)
+		if self._child.Parent == instance then
```
Quenty, line 33: We should probably use self._binder:Observe(instance) on the parent instead for performance. But this works, and the binder is already being listened to for the above area. Rx is definitely preferred here for pref long term.

Quenty (approved):

PR #602, opened 2025-11-04
fix: mock players in studio local server dont have permissions to use cmdr

Daimywil:
When running a studio playtest with multiple players, their userids will be negative and `RunService:IsStudio()` returns false. This fix handles those playtests correctly.


src/permissionprovider/src/Server/Providers/GroupPermissionProvider.lua
```diff
@@ -147,6 +147,12 @@ end
 function GroupPermissionProvider._handlePlayer(self: GroupPermissionProvider, player: Player): ()
 	assert(player, "Bad player")
 
+	-- when testing with mock players in studio, their userids are negative
+	if player.UserId < 0 then
```
Quenty: we need to add a check for `RunService:IsStudio()` since guests are historically considered negative userIds in case Roblox ever does this again.

Daimywil: I was under the impression that returned false in local servers but I was wrong (its just team tests). I'll add that in.

Daimywil: Do you think we should just completely replace the negative userid checks with `IsStudio` checks?

PR #698, opened 2026-05-21
feat: added SetValue method to attributevalue

KreatorKols:


src/attributeutils/src/Shared/AttributeValue.lua
```diff
@@ -1,21 +1,21 @@
 --!nonstrict
 --[=[
-	Allows access to an attribute like a ValueObject.
+    Allows access to an attribute like a ValueObject.
```
Quenty, line 3: aren't these suppose to be tabs?

KreatorKols: No but my vs doesn't even show these as changes I wasn't sure how to resolve

PR #706, opened 2026-05-29
fix: Prevent playback on invalid targets

alex-y-z:
`AnimationSlotPlayer` attempts playback even as targets are being destructed. In cases where the animator has already been destroyed, `HumanoidAnimatorUtils` creates a new one on the client, which produces a warning.

This checks if the animation target has a valid parent before attempting playback.


Quenty:
:rocket: PR was released in [`@quenty/animations@8.32.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/animations%408.32.1) :rocket:

PR #714, opened 2026-06-20
fix: Flush save slot deletions

alex-y-z:
Immediately saves after slot deletion to prevent stale reads.


Quenty:
:rocket: PR was released in [`@quenty/saveslot@1.2.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%401.2.1) :rocket:

PR #718, opened 2026-06-24
feat(chatproviderservice): convert package to --!strict typing

Quenty:
Converts the remaining ChatProviderService package modules from --!nonstrict to --!strict with explicit Nevermore-style type annotations. Along the way it fixes two latent bugs the type checker surfaced: the developer and admin tags were built with a misnamed LocalizedText field (now TagLocalizedText) so their localized text was silently dropped, and ChatTagBase typed ChatTagKey as a boolean even though it stores the string tag key. No public API behavior changes otherwise.

### DIFF
--- a/src/chatproviderservice/src/Client/Binders/HasChatTagsClient.lua
+++ b/src/chatproviderservice/src/Client/Binders/HasChatTagsClient.lua
@@ -1,4 +1,4 @@
---!nonstrict
+--!strict
 --[=[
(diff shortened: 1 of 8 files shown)


Quenty:
:rocket: PR was released in [`@quenty/chatproviderservice@9.53.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/chatproviderservice%409.53.0) :rocket:

PR #719, opened 2026-06-24
feat(deathreport): convert package to --!strict

Quenty:
Converts every non-strict source file in the deathreport package to --!strict, adding explicit type annotations across the death, kill, and team tracker classes, DeathReportUtils, and the client service. Binder registrations hoist their class requires to locals so the sanctioned any cast doesn't trip luau-lsp's require resolver. There are no runtime behavior changes.


Quenty:
:rocket: PR was released in [`@quenty/deathreport@10.45.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/deathreport%4010.45.0) :rocket:

PR #721, opened 2026-06-30
feat: Add server-only purchasing mode to GameProductService

unrooot:
By default a client can prompt a purchase for any asset id via `GameProductServiceClient:PromisePromptPurchase`. When selling limited UGC this lets an exploiter prompt arbitrary asset ids. Enabling server-only prompting makes the client reject local prompts so that prompts must be initiated from the server, where you can validate the asset id:

```lua
-- Server
local gameProductService = serviceBag:GetService(require("GameProductService"))
gameProductService:SetServerOnlyPromptingEnabled(true)

-- Now prompt from the server (validate idOrKey first), e.g.
gameProductService:PromisePlayerPromptPurchase(player, GameConfigAssetTypes.ASSET, assetId)
```

On the client, `GameProductServiceClient:PromisePromptPurchase` rejects while this mode is enabled. Use `GameProductServiceClient:ObserveServerOnlyPromptingEnabled()` to hide or disable buy buttons.

Note: this prevents the helper API from prompting arbitrary ids; a client can still call Roblox's `MarketplaceService:PromptPurchase` directly, so server-side validation of what you sell remains the real authority.

### DIFF
--- a/src/gameproductservice/src/Client/GameProductServiceClient.lua
+++ b/src/gameproductservice/src/Client/GameProductServiceClient.lua
@@ -147,6 +147,17 @@ function GameProductServiceClient:PromisePromptPurchase(
 end
 
 --[=[
+	Observes whether server-only prompting is enabled. When enabled,
+	[GameProductServiceClient:PromisePromptPurchase] rejects and prompts must be
+	initiated from the server. Useful for hiding or disabling buy buttons on the client.
+
+	@return Observable<boolean>
+]=]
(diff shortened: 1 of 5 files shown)


Quenty (approved):

Quenty:
:rocket: PR was released in [`@quenty/gameproductservice@14.52.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/gameproductservice%4014.52.0) :rocket:

PR #722, opened 2026-07-02
fix: Fix attribute replication for server-only prompting

unrooot:
previously was not properly replicating the state of the function we added from #721


Quenty:
:rocket: PR was released in [`@quenty/gameproductservice@14.52.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/gameproductservice%4014.52.0) :rocket:

PR #728, opened 2026-07-14
test(octree): Add unit tests for Octree, OctreeNode, and OctreeRegionUtils

Quenty:
Adds Jest-based unit tests for the octree package covering the public API and internal region utilities, including edge cases.

The tests cover Octree (empty searches, coincident nodes, radius boundaries, k-nearest ordering and bounds, negative and cross-region coordinates, argument validation), OctreeNode (position get/set, no-op reposition, region moves, destruction, unpositioned destroy), and OctreeRegionUtils (inclusive region bounds, octant classification, top-level cell rounding at boundaries and negatives, hashing, region reuse/lookup, subregion creation, node add/move/remove with subregion pruning, and radius neighbor collection).

The package is also wired up for testing following the docs setup: jest.config.lua, a test place project, a server script template, and a test deploy target.

Note: the test deploy target currently has a placeholder placeId. It needs a real place provisioned via `nevermore deploy init --create-place` (this could not run here because the environment ROBLOSECURITY cookie is expired). Until then the cloud tests job will not be able to run the octree tests.


Quenty:
:rocket: PR was released in [`@quenty/octree@11.14.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/octree%4011.14.2) :rocket:

PR #731, opened 2026-07-16
feat(nevermore-claude): add Claude Code plugin with strict-typing-luau skill

Quenty:
Packages the strict-typing-luau skill as a Claude Code plugin so Nevermore and its sister repos can share one install. The skill moves out of `.claude/skills` into `tools/nevermore-claude`, and the monorepo becomes its own plugin marketplace via `.claude-plugin/marketplace.json` at the repo root. Projects scaffolded with `nevermore init` (games and plugins) now ship a `.claude/settings.json` that registers the marketplace and enables the plugin, so collaborators are prompted to install it when they trust the repo. There is no separate publish step — merging to main publishes the plugin, and its version rides the existing lerna/Auto pipeline (mirrored into plugin.json by a small sync script).

The marketplace source uses sparsePaths so teammates check out only the catalog and this plugin instead of cloning the whole monorepo into their plugin cache. The catalog is kept at the repo root deliberately: a source `path` field to a nested catalog exists in the schema but breaks `marketplace update` on Claude Code 2.1.211, so root is required for a working update flow.

Validated end-to-end against this branch on GitHub: marketplace update fetches and validates, the plugin resolves with the strict-typing-luau skill, sparse-checkout trims the clone correctly, and install succeeds.


Quenty:
:rocket: PR was released in [`@quenty/nevermore-cli@4.37.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli%404.37.0), [`nevermore-claude@0.2.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/nevermore-claude%400.2.0) :rocket:

PR #747, opened 2026-07-23
fix(saveslot): selection chain consumes loads that settle after the player leaves

Quenty:
## Problem

The per-player selection chain in `SaveSlotService` (slots loaded → teleport read → default slot) runs against datastore reads that can settle long after the player is gone — session-lock and datastore retries outlive a leave. Only the outermost promise was maid-owned, so a late settle ran continuations that called into the destroyed (metatable-stripped) `HasSaveSlots` binder:

```
SaveSlotService:125: attempt to call missing method 'PromiseLastActiveSlotId' of table
```

egg-hunt-2026's cloud test suite hit this as a run-failing stray error (currently worked around with a `pnpm patch`, which this PR replaces); on a live server it's the same crash-trace noise on every mid-load leave.

## Fix

- Every hop in the chain — including all of `_promiseSelectDefaultSlot` — is now **maid-owned**, so pending reads cancel with the binder's brio.
- Each continuation **re-checks the binder is alive** (`hasSaveSlots.Destroy`) before calling into it: a continuation queued in the same frame the binder dies still executes, so cancellation alone doesn't close the window.

## Testing

New `SaveSlotLateSettle.spec.lua` pins both windows with a yield-injected `DataStoreMock`: unbinding while the slots-load is still pending, and unbinding inside the default-slot read's in-flight window so the settle lands on the dead binder. `nevermore test --cloud` for the saveslot package: **87/87 passed**, no stray errors.

https://claude.ai/code/session_01LQCjTHuCgwir8JZzeAeySJ

### DIFF
--- a/src/saveslot/src/Server/SaveSlotLateSettle.spec.lua
+++ b/src/saveslot/src/Server/SaveSlotLateSettle.spec.lua
@@ -1,0 +1,122 @@
+--!strict
+--[[
+	The per-player selection chain (slots loaded -> teleport read -> default slot) runs against
+	datastore reads that can settle long after the player left — session-lock and datastore
+	retries outlive a leave. A continuation that then calls into the destroyed
+	(metatable-stripped) HasSaveSlots binder throws "attempt to call missing method ..." as a
+	stray error that fails the whole run. The chain must consume a late settle silently: every
+	hop is maid-owned (cancelled with the binder's brio) and re-checks the binder is alive.
+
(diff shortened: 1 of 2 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@1.14.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%401.14.1) :rocket:

PR #749, opened 2026-07-24
fix(saveslot): bind the client HasSaveSlots for a mocked local player

Quenty:
## Problem

`HasSaveSlotsClient.new` gated construction on `player ~= Players.LocalPlayer`:

```lua
if player ~= Players.LocalPlayer then
    return nil :: any
end
```

A `PlayerMock` is a `Folder` and can never be the engine's real `Players.LocalPlayer`, so in a dual-realm test (server bag + client bag + a mock as the local player) the client-realm `HasSaveSlots` Tie was never implemented for the mock. As a result `SaveSlotDataService:ObserveActiveSlotId` — and anything built on it, e.g. a client menu's slot-resume flow — never emitted for the mock, even though the server was writing the shared `ActiveSlotId` attribute on that very instance. The client save-slot flow was effectively untestable headless.

## Fix

Resolve the gate through the `PlayerMock` fallback, the same pattern used across the engine's other client seams (`Players.LocalPlayer or PlayerMock.getMockedLocalPlayer()`):

```lua
if player ~= (Players.LocalPlayer or PlayerMock.getMockedLocalPlayer()) then
    return nil :: any
end
```

No behavior change for real clients — `getMockedLocalPlayer()` returns `nil` when nothing is designated, so the expression collapses to `Players.LocalPlayer`. `@quenty/playermock` is already a dependency of this package.

## Tests

New `HasSaveSlotsClient.spec` (the package's first client-binder / client-realm coverage), verified with `nevermore test --cloud` — **90/90** in the saveslot suite:
- creates **no** client implementation for a mock that isn't the local player (gate rejects → `HasSaveSlotsInterface:Find(mock, CLIENT)` is nil);
- creates the client implementation for the designated local-player mock;
- `SaveSlotDataService:ObserveActiveSlotId` then emits the active slot id and tracks it as the `ActiveSlotId` attribute changes.

stylua, selene, and `luau-lsp analyze` are clean.

### DIFF
--- a/src/saveslot/src/Client/Binders/HasSaveSlotsClient.lua
+++ b/src/saveslot/src/Client/Binders/HasSaveSlotsClient.lua
@@ -10,6 +10,7 @@ local Players = game:GetService("Players")
 local Binder = require("Binder")
 local HasSaveSlotsBase = require("HasSaveSlotsBase")
 local HasSaveSlotsInterface = require("HasSaveSlotsInterface")
+local PlayerMock = require("PlayerMock")
 local Promise = require("Promise")
 local Remoting = require("Remoting")
 local SaveSlotConstants = require("SaveSlotConstants")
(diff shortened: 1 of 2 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@1.14.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%401.14.2) :rocket:

PR #754, opened 2026-07-26
fix(teleportserviceutils): report a refused teleport instead of raising

Quenty:
A client teleport the engine refuses outright no longer throws out of promiseTeleportClient and tears down the calling thread. Studio refuses every teleport this way, so the error surfaced as an uncaught crash in the caller even when it handled the promise rejection.

TeleportServiceUtils.teleport now pcalls, returning whether the teleport started plus the reason it did not, and promiseTeleportClient acts on that return: a raise is treated as the same failure TeleportInitFailed reports, so it retries and rejects once the attempts are spent rather than escaping. A refused hop reports itself both ways, so only the first report of an attempt counts.

Tests cover the raised-refusal path; reverting the fix fails them with the same "Cannot Teleport"-shaped trace out of promiseTeleportClient as the original report.

https://claude.ai/code/session_01MN44KnxAF829kJwZRKUMef

### DIFF
--- a/src/teleportserviceutils/src/Shared/TeleportServiceUtils.lua
+++ b/src/teleportserviceutils/src/Shared/TeleportServiceUtils.lua
@@ -1,20 +1,15 @@
 --!strict
 --[=[
-	Utilities for teleporting players, including mock-aware wrappers of the user-facing TeleportService
-	teleport APIs. With a [PlayerMock] (a headless test), each teleport is recorded on the mock -- the
-	`"TeleportService.Teleport"` lookup domain, keyed by destination placeId -- instead of calling the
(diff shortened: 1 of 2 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@2.2.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%402.2.2), [`@quenty/softshutdown@10.2.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/softshutdown%4010.2.2), [`@quenty/teleportserviceutils@10.2.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/teleportserviceutils%4010.2.2) :rocket:

PR #756, opened 2026-07-27
feat(saveslot): save slot commands take a players argument

Quenty:
Every Cmdr command in the save slot package now takes a player argument ahead of its existing arguments, the same shape export-save-slot already had, so an admin can act on other players instead of only themselves. Previously only the two export commands could. Each command reports its result per player.

The two import commands take a single Player, since applying one code across everyone is disruptive enough to be worth spelling out per player, and the ephemeral import also selects the slot it creates. The rest take Players.

The export commands can now also export a player's main slot, which the normal path refuses. That refusal exists because the main slot's store is the player's shared root datastore, so reading it pulls the SaveSlots system data alongside the slot's own data, meaning the roster and every other slot. A main slot export now strips that key the same way slot duplication already did, so the export carries only the slot itself and importing it cannot nest a roster inside a slot. Universe-scoped global data sharing the root store cannot be told apart from the main slot's own data and is still carried, so the opt-in is deliberately not available through SaveSlotService, only to code holding the binder directly. The export methods were already server realm only, so there is no client path to this.

Two commands changed beyond gaining the argument. reset-save-slot's Slots argument becomes required, because with Players ahead of it an optional Slots would leave `reset-save-slot *` as valid syntax meaning "wipe the active slot of every player in the server", where it used to mean "reset all of my slots". And create-save-slot was reading the MaxSlotCount cap off the executor, which would have validated every target against the executor's cap once it took players, so it now reads each player's own cap from the binder.

Note that the slot `.` and `*` operators still resolve against the executor's slot list, since Cmdr types only ever see the executor. For the commands taking both players and slot indices, that means "each player's slots at the indices I have", not "each player's own slots". That was already true of the export commands.

https://claude.ai/code/session_014vxqfL1jx7niZx3xKUaCTG

### DIFF
--- a/src/saveslot/src/Server/Binders/HasSaveSlots.ExportImport.spec.lua
+++ b/src/saveslot/src/Server/Binders/HasSaveSlots.ExportImport.spec.lua
@@ -158,6 +158,39 @@ describe("HasSaveSlots.PromiseExportSlot / PromiseImportSlot", function()
 		end)
 	end)
 
+	it("exports the main slot when allowMainSlot is set", function()
+		runWithContext(function(context)
+			local hasSaveSlots = context.hasSaveSlots
+
+			local mainSlotId = createSelectAndWrite(hasSaveSlots, SaveSlotConstants.DEFAULT_SLOT_INDEX)
+
+			local export = awaitValueOf(hasSaveSlots:PromiseExportSlot(mainSlotId, true))
+			expect(export.data.Coins).toEqual(7)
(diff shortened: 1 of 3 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@2.3.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%402.3.0) :rocket:

PR #757, opened 2026-07-27
fix(settings): stop settings hydration recursing until the C stack dies

Quenty:
## The bug

A client crash during teleport teardown, bottoming out in a wall of nested `EventHandlerUtils` frames.

`SettingsDataService` filled its per-player hydration memo *after* `_hydrateCacheForPlayer` returned:

```lua
self._hydratedPlayersMaid[player] = self:_hydrateCacheForPlayer(player)
```

Hydration subscribes and emits synchronously, so the slot stayed empty for that whole emission. Any settings read arriving during it — an observer that reads a setting while handling one, which is ordinary — hydrated again. Because `TieDefinition._observeImplementedInterface` mints a fresh `TieInterface` per subscription, `ObservableMap.Set`'s identity dedupe never engaged, so the map re-emitted and fed the next read. Nested signal fires resume on the C stack (`EventHandlerUtils.fire` clears the free thread *before* running the handler), so this recursed until the stack died.

`Players.PlayerRemoving` was the trigger: it cleared the memo while observers were still subscribed, and the teardown emission itself carried the read that re-entered.

## The fix

- Claim the slot **before** hydrating, bounding it to one hydration.
- Tombstone a departing player into that same slot instead of clearing it, so the read that arrives during teardown finds it claimed. The tombstone drops itself on `AncestryChanged` once the player is out of the DataModel, and an ancestry check keeps a departed player from being hydrated or held afterwards.
- Key the cache off the player being hydrated rather than calling the implementation's `GetPlayer` across the tie — a bindable round-trip per emission, and a nil key if the folder reparents.
- Assert on re-entrant hydration, so this class of bug errors legibly instead of crashing the process.

Second commit reduces the work that made the loop expensive: `GetSettingProperty` caches per `(serviceBag, player)`, `SettingProperty.Observe` caches its pipeline and drops re-emissions from tie churn that never changed the value, and `SetValue` returns early on a write that changes nothing.

## Verification

`nevermore test --cloud` in `src/settings`: **14/14** (was 5).

The regression specs were checked against the broken code rather than only against the fix:

| Reverted | Result |
|---|---|
| slot claimed after hydrating | `impl,nil,impl,impl,…` — 94 emissions and a stack blowout in `_initializeThread` |
| tombstone replaced with clearing the slot | re-hydrates during teardown; departed player still held |
| `Rx.distinct()` removed | 2 tie re-implementations → 5 emissions instead of 1 |

Also green downstream: the consuming game's suite is unchanged at 340/367 (27 pre-existing skips, zero failures). stylua, selene and luau-lsp clean.

## Note for reviewers

`PlayerSettingsClient.SetValue`'s change is untested. The class gates its entire client branch on `self:GetPlayer() == Players.LocalPlayer` — a literal identity check that is nil headless — so `_toReplicateCallbacks` never initializes and it cannot be exercised in a cloud test. Every neighbour (`PlayerSettingsBase.GetPlayer`, `SettingProperty._findPlayer`, `SettingDefinition`) already has a `PlayerMock` seam; this is the holdout. Happy to add one in a follow-up if you want that coverage.

### DIFF
--- a/src/settings/src/Client/Player/PlayerSettingsClient.lua
+++ b/src/settings/src/Client/Player/PlayerSettingsClient.lua
@@ -224,6 +224,13 @@ function PlayerSettingsClient.SetValue<T>(self: PlayerSettingsClient, settingNam
 		end
 	end
 
+	-- A reader already sees this value -- queued, in transit, or replicated. Observers dedupe on value
+	-- themselves; returning rather than just skipping the notification is what also skips the
+	-- redundant round-trip to the server.
+	if self:_getCurrentValue(settingName) == value then
+		return
+	end
+
 	local queueReplication = false
 	local toReplicate
 	if self._toReplicate then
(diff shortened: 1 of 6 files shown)


Quenty:
:rocket: PR was released in [`@quenty/settings-inputkeymap@10.69.3`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/settings-inputkeymap%4010.69.3), [`@quenty/settings@11.66.3`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/settings%4011.66.3) :rocket:

PR #758, opened 2026-07-27
feat(saveslot): run consumer callbacks before a slot is selected

Quenty:
## Why

Every way a save slot can become active funnels through one commit in `HasSaveSlots.PromiseSelectSlot`: an explicit selection, a new or ephemeral slot, the default slot on join, an arrival resuming the slot it teleported in with, and Cmdr. A consumer with per-selection state to settle *before* the slot changes had nowhere to put it but each of those entry points individually — and missing one leaves state from the previous selection standing over the new slot.

That is a live bug in a game built on this package: a "spawn in this place" stake made for a throwaway ephemeral slot survived into the *next* slot the player picked, because the picker's selection path never passed the code that owned the stake.

## What

`SaveSlotService:RegisterPreSelectCallback(callback)` — one game-wide registry, applying to every player, bound or not yet.

A callback runs with the previous selection still in place, receiving `(player, slotId, previousSlotId)`, and its return value decides what happens next:

| Returns | Effect |
| --- | --- |
| nothing / `true` | allow the selection |
| `false` | refuse it — `PromiseSelectSlot` rejects, active slot untouched |
| a promise | hold the selection open until it settles; resolving `false` refuses |

An error, or a rejected promise, is isolated and warned about rather than read as a refusal — a refusal is a decision a callback states, never one inferred from it crashing, and one consumer's bug must not be able to stop every player in the game from loading a save slot. This mirrors the call the summary providers already make with their `NONE` sentinel. For the same reason there is no timeout: a callback that never settles holds the selection, so consumers keep the work bounded.

`HasSaveSlots` reads the registry from the service as each selection commits, through a private accessor, rather than keeping a mirrored copy per player. The accessor answers empty when no `SaveSlotService` is in the bag — a binder can be bound on its own, and nobody could have registered through a service that isn't there — and requires the service lazily, since `SaveSlotService` requires the binder.

## Testing

`nevermore test --cloud` in `src/saveslot`: **180/180 pass** (162 before this change). `lint:luau` and `selene` clean.

`HasSaveSlots.PreSelect.spec.lua` covers ordering (the callback sees the slot it is replacing, not the one arriving), every selection entry point, the no-op re-selection, removal, the async hold, refusal by value and by promise, and error isolation. It stubs the private accessor to inject callbacks: booting `SaveSlotService` into that bag drags a second `PlayerMockService` into a DataModel another suite is already using one in. `SaveSlotService.spec.lua` covers the registry itself — registration, removal, post-Start registration, and that readers get a snapshot rather than the live list.

The seam between the two (a service registration actually reaching a binder's fan-out) is exercised where both are booted together, in the consuming game's specs.

## Notes

- Two Luau papercuts worth knowing about: parameterizing `ObservableSet` by a function type walks `HasSaveSlots.lua` into the "code is too complex to typecheck" ceiling (so the set is stored as `any` with the intended type in a comment), and the lazily-required service reads as a distinct type from the one `GetService`'s generic binds ("Expected SaveSlotService, got SaveSlotService"), so that handle is typed `any`.
- Under `--!strict`, a side-effect-only callback needs an explicit `return nil`, since Luau requires all codepaths to return once the return type is non-void. That is the cost of typing the veto contract; happy to make the type permissive instead if you'd rather.

https://claude.ai/code/session_01JE2YH54vfc5mbmDMUH4XoU

### DIFF
--- a/src/saveslot/src/Server/Binders/HasSaveSlots.PreSelect.spec.lua
+++ b/src/saveslot/src/Server/Binders/HasSaveSlots.PreSelect.spec.lua
@@ -1,0 +1,382 @@
+--!strict
+--[[
+	Coverage for the pre-select callback: the seam a consumer uses to settle per-selection state before a
+	slot becomes active, whatever selected it.
+
+	@class HasSaveSlots.PreSelect.spec.lua
+]]
+local require = require(script.Parent.loader).load(script)
+
+local DataStoreMock = require("DataStoreMock")
+local Jest = require("Jest")
+local PlayerDataStoreService = require("PlayerDataStoreService")
+local PlayerMock = require("PlayerMock")
+local Promise = require("Promise")
+local PromiseTestUtils = require("PromiseTestUtils")
+local ServiceBag = require("ServiceBag")
+
+local Workspace = game:GetService("Workspace")
+
+local afterEach = Jest.Globals.afterEach
+local describe = Jest.Globals.describe
+local expect = Jest.Globals.expect
+local it = Jest.Globals.it
+
+local FAKE_USER_ID = 424243
+
+local activeContext: any = nil
+
+-- A test that fails mid-body never reaches its own destroy, and the PlayerMock it leaves parented to
+-- Workspace is consumed by the next suite that boots a PlayerMockService -- which fails *that* suite
+-- instead of this one. destroy is idempotent, so this is a no-op after a test that tore itself down.
+afterEach(function()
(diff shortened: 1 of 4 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@2.4.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%402.4.0) :rocket:

PR #759, opened 2026-07-27
Users/quenty/access

Quenty:

PR #761, opened 2026-07-28
fix(access): unregister a fact, feature or policy when it is destroyed

Quenty:
## What

A destroyed `AccessFact` stayed in the registry, so the merge still called methods on it. `BaseObject.Destroy` nils the metatable, so those calls don't return nothing — they fail:

```
AccessDataService:1097: attempt to call missing method 'GetServerOverrideBehavior' of table
```

Caught in egg-hunt CI, where it produced **10 tracebacks and zero failed assertions** — the harness reads any traceback as a failure, so the job went red with every test passing.

## Why it happens

Fallout from making these objects maid-owned in #759. One maid commonly holds both the object and its unregister disposer, and `Maid.DoCleaning` runs its tasks in no particular order — so relying on the caller to unregister before the object dies was relying on luck.

## The fix

Where the invariant belongs, rather than a nil-guard at the call site: `RegisterFact`, `RegisterFeature` and `RegisterPolicy` hook the object's **own** maid. `BaseObject.Destroy` runs that maid *before* nilling the metatable, so the object leaves the registry while it is still a usable object — a moment later, every method on it would fail instead. The disposers are idempotent, so the ordinary path calling them too is fine, in either order.

## Verification

`nevermore test --cloud` in `src/access` — 285/285. Four new tests: a destroyed fact leaves the registry, a live report survives one being destroyed under it, the same for features, and calling the disposer as well is harmless. luau-lsp 0 errors, stylua and selene clean.

Unblocks egg-hunt PR #315. Its other failure was a `PlayerMock` read in Raven's `owner` package, fixed separately in QuentyStudios/Raven#663.

### DIFF
--- a/src/access/src/Shared/AccessDataService.lua
+++ b/src/access/src/Shared/AccessDataService.lua
@@ -653,7 +653,13 @@ function AccessDataService.RegisterFact(self: AccessDataService, fact: AccessFac
 	end)
 	self:_refreshFactNames()
 
-	return function()
+	local unregistered = false
+	local function unregister()
+		if unregistered then
+			return
+		end
+		unregistered = true
+
 		local current = self._layersByFactName[factName]
 		if not current then
 			return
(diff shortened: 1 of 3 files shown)


Quenty:
:rocket: PR was released in [`@quenty/access@1.1.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.1.1) :rocket:

PR #762, opened 2026-07-28
fix(access): keep a destroyed layer out of a merge already in flight

Quenty:
Follow-up to #761, which fixed the common case but not the one still failing in egg-hunt CI. That run went from 45 tracebacks to 9 — the Raven half cleared, this one did not:

```
AccessDataService:1124: attempt to call missing method 'GetServerOverrideBehavior' of table
```

## What was still wrong

**The disposer looked the fact up through the registry.** The registry and the array a subscribed report holds come apart in two ways: `Destroy` clears the registry table, and re-registering a name after its last layer left builds a fresh array. Either way `self._layersByFactName[factName]` no longer points at the array the report closure captured — so the lookup finds nothing, returns early, and the dead fact sits in that array until the next emission calls a method on it. By then `BaseObject.Destroy` has nilled its metatable, so the method is *missing* rather than merely unhelpful. It now removes from the array it inserted into.

**The merge now reads only live layers.** `AccessFact.isAccessFact` is false for a destroyed fact, so a dead one stops contributing instead of taking the merge down. That's the belt, not the braces — teardown order isn't ours to control, and a merge in flight shouldn't be takeable down by a layer that went while it was running.

## What is and isn't proven

`nevermore test --cloud` — 287/287, luau-lsp 0 errors, stylua and selene clean. Two new tests cover a layer destroyed under a live report: the merge keeps working, and the readout drops the row rather than showing a dead one.

**Being straight about the limit:** I could not reproduce the exact CI sequence in a unit test. I wrote three attempts that passed with the fix fully reverted, so I threw them away rather than ship green-looking cover. The tests here assert real behaviour but would not have caught this specific bug. The proof will be egg-hunt #315 going from 9 tracebacks to 0 — if it doesn't, this diagnosis is wrong and I'd rather find that out from CI than from a test I talked myself into.

### DIFF
--- a/src/access/src/Shared/AccessDataService.lua
+++ b/src/access/src/Shared/AccessDataService.lua
@@ -660,17 +660,20 @@ function AccessDataService.RegisterFact(self: AccessDataService, fact: AccessFac
 		end
 		unregistered = true
 
-		local current = self._layersByFactName[factName]
-		if not current then
-			return
-		end
-
-		local index = table.find(current, fact)
+		-- The array this fact was inserted into, not whatever the registry currently points at. Destroy
+		-- clears the registry table, but a live report closure still holds the array -- so going through
+		-- the registry would find nothing, leave the fact in that array, and let the next emission call a
+		-- method on an object whose metatable is about to be nil.
+		-- The array this fact was inserted into, not whatever the registry currently points at. The two come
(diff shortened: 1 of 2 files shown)


Quenty:
:rocket: PR was released in [`@quenty/access@1.1.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.1.2) :rocket:

PR #763, opened 2026-07-28
fix(access): make client-realm policies actually run, and give them an off switch

Quenty:
Two changes to `AccessPolicyService`, both about the same thing: a policy whose consequence runs on the client was unusable.

## 1. `fix`: run policies against the bag's tie realm, not RunService

`AccessPolicyService` decided where a policy runs by asking `RunService:IsServer()`. Every other realm branch in this package reads `TieRealmService`, and `AccessDataService` says why in as many words:

> Branching on the tie realm rather than on RunService, for the same reason the tie itself does: it is the one realm answer a service bag can be told, which is what lets both halves be booted together and actually exercised against each other.

The two come apart wherever a bag is *told* its realm — a test booting both halves in one DataModel is the whole reason `TieRealmService` exists. In that setup a `CLIENT`-realm policy never ran.

The failure was silent rather than loud: the policy is registered, appears in `GetPolicyNames`, reports `true` from `IsPolicyEnabled`, shows up in `GetDebugState` — and does nothing. `IsPolicyActiveForPlayer` is the one method that would have contradicted that, and it read `RunService` too, so the readout agreed with the bug.

Both call sites now go through one `_runsHere`. Anything not explicitly `TieRealms.CLIENT` counts as the server, matching how `AccessDataService` branches, so a shared bag still runs server-realm enforcement. No behaviour change in a real client or a real server, where the two agree.

## 2. `feat`: replicate policy enablement from server to client

Once a client-realm policy runs, it still has no off switch. `access-policy <name> off` is a server-side command (`AccessCommandService` is `@server`), `AccessPolicyServiceInterface` is query-only on purpose, and nothing joined the two — so switching off a client-realm policy did nothing, silently, with both realms' readouts agreeing it was off.

That matters most for what policies are *for*: taking one layer of enforcement out of the way to prove the layer beneath it catches what the first one missed. A layer you cannot disable cannot be tested around.

Enablement now replicates one way, server → client, as state on `ReplicatedStorage` — deliberately **not** a new tie method, so the interface's rule holds: finding the registry is still not permission to start enforcing.

Two details worth reviewing:

- **The whole effective map is published, including the policies that are off.** A client whose own `isEnabledByDefault` switched something on has to hear that the server says otherwise, and "absent from the payload" cannot carry that.
- **A policy the server never registered is left alone.** Replication widens what the client knows; it does not own the client's registry. Same rule as `_reconcileReplicatedFeatureFacts`.

`SetReplicatedPolicyEnabled` is exposed as the seam a test drives directly, matching `AccessDataService.SetReplicatedFactOverrides`.

## Tests

`nevermore test --cloud`: **292/292** (was 287).

`AccessPolicyService.spec` — an optional `tieRealm` argument on `setup()` so a client bag can be built at all:
- a client-realm policy **runs** in a bag told it is a client, and `IsPolicyActiveForPlayer` agrees
- a server-realm policy in that same bag still does **not** run

`AccessRealmReplication.spec` — both halves booted, exercised through the real transport:
- a policy switched **off** on the server reaches the client
- a policy switched **on** on the server reaches the client
- a policy the server never registered keeps what it was built with

## Found by

Building the enforcement layer for a game: a client-realm "hold the paywall up while this player may not play here" policy. It could not be exercised from the test runner (1), and could not be switched off to test the server-side layer beneath it (2).

### DIFF
--- a/src/access/src/Shared/AccessPolicyService.lua
+++ b/src/access/src/Shared/AccessPolicyService.lua
@@ -129,26 +129,20 @@ function AccessPolicyService.Start(self: AccessPolicyService): ()
 end
 
 --[[
-	Publishes which policies are switched on, so a console session on the server reaches the realm the
-	policy actually runs in.
+	Publishes which policies are switched on, so a server-side console reaches the realm a policy runs in.
 
-	This is the half that was missing. `AccessPolicyServiceInterface` is query-only on purpose -- letting
-	anything that merely found an interface start a kick would make "can see the registry" and "can
-	enforce" the same permission -- but that left a client-realm policy with no off switch at all: the
-	command is server-side, the consequence is client-side, and nothing joined them. So enablement
-	replicates one way, server to client, as state rather than as a callable.
+	State rather than a tie method: [AccessPolicyServiceInterface] stays query-only, so finding the
+	registry is not permission to start enforcing.
 
-	The whole effective map is published rather than the changes to it, and includes the policies that are
-	*off*: a client whose own `isEnabledByDefault` switched something on has to hear that the server says
-	otherwise, and "absent from the payload" cannot carry that.
+	The whole effective map is published, including the policies that are *off* -- a client whose own
+	`isEnabledByDefault` switched something on has to hear that the server says otherwise, and
+	absent-from-the-payload cannot carry that.
 ]]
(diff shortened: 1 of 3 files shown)


Quenty:
:rocket: PR was released in [`@quenty/access@1.2.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.2.0) :rocket:

PR #764, opened 2026-07-28
feat(saveslot): let save-slot commands address ephemeral slots and persist them

Quenty:
The save-slot Cmdr commands could not name an ephemeral slot, since ephemeral slots are kept out of the slot list, so a throwaway session could not be exported, ended, or turned into a save. The slot-index type now offers the reserved ephemeral index while a session is active and the commands resolve it against the active slot, so "." and an explicit 0 both reach it.

Duplicating an ephemeral slot is now allowed and copies its live in-memory data into a real slot, and the new persist-save-slot command does that and switches onto the copy so play continues on data that actually saves. Deleting an ephemeral slot ends the session instead of being refused for being the active slot, and list-save-slots shows the active session.

The slot-index type also accepts any literal index the executor has no slot for, so you can act on another player's slot that your own autocomplete cannot see.

All 188 saveslot tests pass on cloud.

### DIFF
--- a/src/saveslot/src/Server/Binders/HasSaveSlots.lua
+++ b/src/saveslot/src/Server/Binders/HasSaveSlots.lua
@@ -40,13 +40,6 @@ export type SummaryProvider = (Player, DataStoreStage.DataStoreStage) -> Observa
 -- identity, so it never collides with a real value (even an empty table) a provider might emit.
 local NONE = {}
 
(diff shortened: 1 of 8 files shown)


Quenty:
:rocket: PR was released in [`@quenty/saveslot@2.6.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%402.6.0) :rocket:

PR #765, opened 2026-07-28
feat(access): toggle several policies at once with access-policy

Quenty:
The access-policy console command now takes a list of policies instead of a single one, so `access-policy * off` (or `all`, or a comma-separated list) switches every registered consequence off in one go, matching what access-override already did for facts. Names that aren't registered are reported before anything is toggled, so a typo in the middle of a list doesn't leave half of it switched.


Quenty:
:rocket: PR was released in [`@quenty/access@1.3.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.3.0) :rocket:

PR #767, opened 2026-07-28
fix(access): make the access-policy list readout deterministic and tested

Quenty:
Follow-up to #765. The confirmation from `access-policy` named policies in whatever order Cmdr handed them back, which is hash order rather than anything typed, so it now sorts them. The command handler itself had no test — the promise that a typo in a list leaves nothing switched was asserted only by the commit message — so there is now a spec that reaches the registered commands directly. Two existing assertions in the utils spec were passing for the wrong reason and are corrected, and the error and confirmation strings read properly for a list.


Quenty:
:rocket: PR was released in [`@quenty/access@1.3.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.3.1) :rocket:

PR #768, opened 2026-07-28
fix(access): run client-realm policies for the local player

Quenty:
## The bug

`AccessPolicyService` only applies policies to players handed to `AddPlayer`, and the only caller was the server's `AccessService` (`Players.PlayerAdded` → `AddPlayer`). `AccessServiceClient` — its mirror — never added anybody, so `_playerMaids` stayed empty on the client and **no client-realm policy ever ran in a real game**: registered, enabled, listed by `access-policies`, silent.

Found in a game where a client-realm paywall policy never raised its screen while the server-realm policy reading the same feature enforced normally.

Every existing test passed because they all call `AddPlayer` themselves — the wiring between the entry point and the service was the one thing untested.

## The fix

`AccessServiceClient.Start` adds the local player to the policy service, maid-owned.

The local player and nobody else: player instances replicate, so this realm can answer questions about everyone here, but a client-realm consequence is presentation *for the person at this client* — adding the rest would raise this client's screen on a stranger's verdict.

Read once, matching production: `Players.LocalPlayer` exists before any LocalScript runs, and a headless test designates its `PlayerMock` before booting bags for the same reason. A realm with neither runs no client policies.

## Tests

New `AccessServiceClient.spec.lua` boots the client entry point the way a game does: the regression, local-player-only scoping, server-realm skip, and booting with no local player.

Verified both directions with `nevermore test --cloud` in `src/access`: 311/311 green with the fix, and exactly one test fails with the fix reverted.

https://claude.ai/code/session_01FYFnBcgwPT624hfK9rw3GE

### DIFF
--- a/src/access/src/Client/AccessServiceClient.lua
+++ b/src/access/src/Client/AccessServiceClient.lua
@@ -17,11 +17,14 @@
 
 local require = require(script.Parent.loader).load(script)
 
+local Players = game:GetService("Players")
+
 local AccessCommandServiceClient = require("AccessCommandServiceClient")
 local AccessDataService = require("AccessDataService")
 local AccessPlayerClient = require("AccessPlayerClient")
 local AccessPolicyService = require("AccessPolicyService")
 local Maid = require("Maid")
+local PlayerMock = require("PlayerMock")
 local ServiceBag = require("ServiceBag")
 
 local AccessServiceClient = {}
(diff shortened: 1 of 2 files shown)


Quenty:
:rocket: PR was released in [`@quenty/access@1.3.3`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.3.3) :rocket:

PR #769, opened 2026-07-28
fix(receiptprocessing): unregister a receipt processor instead of erroring

Quenty:
## The unregister closure could never unregister anything

`RegisterReceiptProcessor` returns a closure meant to remove the processor. It
indexed `self._handles` — a field the service never creates:

```lua
return function()
	local index = table.find(self._handles, data)  -- self._handles is always nil
	...
end
```

So calling it threw `invalid argument #1 to 'find' (table expected, got nil)`
instead of unregistering. Two consequences:

1. A processor could not be removed for the lifetime of the service.
2. The usual way to hold that closure is `maid:GiveTask(unregister)` — so the
   throw lands during teardown and takes the rest of the maid's tasks with it.

Points it at `_processors`, where the processor actually lives, and returns early
once `Destroy` has cleared the list. That second guard is not defensive padding:
a maid runs its tasks *after* `Destroy` nils `_processors`, which is precisely
when this closure gets called.

## Tests

The package had no specs, so this adds a test target for it following
`docs/testing/testing.md` — `deploy.nevermore.json` (via `nevermore deploy init`),
`test/default.project.json`, the `ServerMain` script template, `jest.config.lua`,
and the `jest-lua` / `nevermore-test-runner` dependencies.

Four cases: the chain routes to a registered processor, the returned unregister
stops routing, it unregisters only its own processor, and it is safe to call after
`Destroy`.

Not a placebo — verified by reintroducing the `_handles` lookup:

| | with `_handles` (old) | with this fix |
|---|---|---|
| routes receipts to the registered processor | ✅ | ✅ |
| stops routing once unregister is called | ❌ | ✅ |
| unregisters only its own processor | ❌ | ✅ |
| survives being called after Destroy | ❌ | ✅ |

## Verification

`nevermore test --cloud` from `src/receiptprocessing`: 4/4 passed.
`nevermore batch test --cloud` (dependents with test targets): `@quenty/access`
311/311, `@quenty/gameproductservice` 80/80.

https://claude.ai/code/session_01KPgnohqfGZQQwWv6PnToJC

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4212,6 +4212,9 @@ importers:
       '@quenty/maid':
         specifier: workspace:*
         version: link:../maid
+      '@quenty/nevermore-test-runner':
+        specifier: workspace:*
+        version: link:../nevermore-test-runner
       '@quenty/playermock':
         specifier: workspace:*
         version: link:../player-mock
@@ -4230,6 +4233,9 @@ importers:
       '@quenty/valueobject':
         specifier: workspace:*
         version: link:../valueobject
+      '@quentystudios/jest-lua':
+        specifier: 3.10.0-quenty.2
+        version: 3.10.0-quenty.2
 
   src/rectutils:
     dependencies:
--- a/src/receiptprocessing/deploy.nevermore.json
+++ b/src/receiptprocessing/deploy.nevermore.json
@@ -1,0 +1,10 @@
+{
+  "targets": {
+    "test": {
+      "universeId": 9716264427,
+      "placeId": 130126340232883,
+      "project": "test/default.project.json",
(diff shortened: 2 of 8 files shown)


Quenty:
:rocket: PR was released in [`@quenty/access@1.3.4`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.3.4), [`@quenty/gameproductservice@14.66.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/gameproductservice%4014.66.2), [`@quenty/receiptprocessing@7.42.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/receiptprocessing%407.42.1) :rocket:

PR #771, opened 2026-07-28
fix(cli): stop reporting a test pass that no test output backs

Quenty:
## Why

A batch run whose engine logs went missing still reported `Passed`. With the per-package log section empty, the jest-failure check is skipped entirely, so the verdict falls through to the Luau `pcall` exit status — "the script didn't throw", not "tests passed". The PR comment then rendered:

```
| egghunt2026 | ✅ Passed (4m11s) |
**1 tested, 1 passed, 0 failed** in 4m20s
```

Those are *package* counts in test-count phrasing, with no test numbers behind them. Nothing in that output distinguishes it from a real result, so a green check proved nothing.

## What this does

Reports what is actually known. **No verdict logic changes — nothing green today turns red.**

- A successful test result carrying no counts renders as **Unverified**, not as a pass. `isEmptyTestRun` only caught "runner said 0 tests"; "we never heard any counts" fell through as success — the same absence-of-evidence bug as the parser, one layer up in reporting.
- Summaries read final results rather than live progress. Progress is cleared once a package resolves, so the existing end-of-run empty check could never fire.
- `1 package tested, 1 passed, 0 failed` — package counts stop reading as test counts.
- Warn when a batch summary parses but no per-package sections do. The summary prints last and survives a truncated fetch; the existing check required *both* to be missing, so the observed failure mode was silent.
- Count Luau tracebacks per package. Jest structurally cannot: a deferred-callback crash fires outside any test, often after the test that scheduled it already passed, so a run prints `Tests: 155 passed, 0 failed` while genuinely broken.
- Record page/message/character counts for fetched task logs, to settle whether pagination is what drops engine output.
- `(no output)` warns and states the result is unverified, instead of printing as benign info.
- `NO_COLOR=1` strips jest-lua's escape codes from echoed engine logs. These come from inside Roblox and are echoed as log content, so chalk never saw them — `NO_COLOR` previously had no effect on the part people actually grep.

## Testing

`vitest run` in both packages: 100 passed in `nevermore-cli`, 63 in `cli-output-helpers`. New coverage for the lost-section case, traceback counting alongside a clean jest summary, and the unverified-render path.

## Follow-ups (not in this PR)

Verdict changes are deliberately held back until this lands and a real run shows how often the unverified state occurs: requiring positive evidence to pass, failing on empty logs, unifying the two divergent parsers (`test-log-parser` fails on `Stack Begin`, `batch-log-parser` ignores it), and attributing tracebacks to their owning package by script path.

https://claude.ai/code/session_014KaJaDsPyPWQ3whHaav2r8

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -155,6 +155,9 @@ nevermore test --cloud --script-text 'print("hi")'    # run arbitrary Luau to de
 | `--output <file>` | Write JSON results to a file. |
 | `--timeout <seconds>` | Max execution time, sent to Open Cloud so Roblox cancels server-side on overrun (default: 120). |
 
+Engine logs are echoed with jest-lua's own colour codes, which the CLI's formatter
+never sees. Set `NO_COLOR=1` to strip them when you intend to grep the output.
+
 ### `nevermore deploy`
 
 Builds a Rojo project and uploads it to a Roblox place. `run` is the default subcommand, so `nevermore deploy` and `nevermore deploy <target>` both deploy. Full walkthrough in [Deploying with the CLI](deploy.md).
--- a/tools/cli-output-helpers/src/reporting/github/formatting.test.ts
+++ b/tools/cli-output-helpers/src/reporting/github/formatting.test.ts
@@ -17,7 +17,12 @@ describe('formatResultStatus', () => {
   it('shows a pass with its test counts', () => {
     const status = formatResultStatus(
       buildResult({
-        progressSummary: { kind: 'test-counts', passed: 275, failed: 0, total: 275 },
+        progressSummary: {
+          kind: 'test-counts',
+          passed: 275,
+          failed: 0,
+          total: 275,
+        },
       }),
       'Passed',
       'Failed',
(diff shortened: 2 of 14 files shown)


Quenty:
:rocket: PR was released in [`@quenty/cli-output-helpers@1.17.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/cli-output-helpers%401.17.1), [`@quenty/nevermore-cli-helpers@1.16.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli-helpers%401.16.1), [`@quenty/nevermore-cli@4.41.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli%404.41.1), [`@quenty/nevermore-template-helpers@1.18.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-template-helpers%401.18.1), [`@quenty/studio-bridge@0.17.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/studio-bridge%400.17.1) :rocket:

Issue #17, opened 2015-11-07
Lazy load modules


JetCrusherTorpedo:
Here is a suggestion:

* Store all modules in ServerStorage.
* Require all modules server-side and store the objects returned by the modules in a table.
* Clients request modules via a RemoteFunction, and the server returns the pre-compiled modules from the table mentioned above.
* Modules are memoized by the client when they are requested for the first time, eliminating the need for the module to be downloaded again, should the client need to require it more than once.

Memoization library: https://github.com/kikito/memoize.lua

Quenty:
Good idea. Technically unfeasible.

ROBLOX RemoteFunctions cannot pass closures, functions, scopes, or metamethods for technical reasons, making this idea, although good, impossible to execute.

Issue #16, opened 2014-12-09
Duplicated code in qGUI and ScreenSpace


Quenty:
There is duplicated code in qGUI and ScreenSpace.

Quenty:
qGUI is deprecated, and dependency of it will be cleaned up later.

Issue #14, opened 2014-10-15
LensGlare Not Operational


learntrblx:
LensGlare fails to create effect
![screenshot_36](https://cloud.githubusercontent.com/assets/9259582/4655556/652b1fd2-54c5-11e4-84cf-227bc9bfdaef.png)
EDIT: Resolved

Tlennyz:
Hi

Issue #13, opened 2014-10-15
Signal not valid Resource


learntrblx:
18:40:53.431 - NevermoreEngine:489: [NevermoreEngineSolo] - The resource 'Signal' does not exist, cannot load
![screenshot_34](https://cloud.githubusercontent.com/assets/9259582/4655521/e49374b4-54c4-11e4-9405-481baaeb0f06.png)
EDIT: Resolved

PR #772, opened 2026-07-29
feat(cli): make a passing test run mean tests were seen to pass

Quenty:
Phase 2, following #771. Phase 1 made an unreadable run *say* so; this makes the verdict follow.

## Root cause, finally pinned

Two 20,000-line probe runs against the real API:

| run | messages | chars | survived |
|---|---|---|---|
| probe 1 (~20 char lines) | 6,715 | ~134,300 | lines 13286–20000 |
| probe 2 (~99 char lines) | 3,237 | 297,722 | lines 16765–20000 |
| egg-hunt | 2,531 | 330,778 | tail only |

**Open Cloud retains only the tail of a task's log.** Both probes lost their head and kept one contiguous block through the final line. `===BATCH_TEST_BEGIN===` prints in the first moments of a run, so on any suite large enough to overflow the window it is *always* the first casualty — which is why this reproduced identically every time rather than flaking.

Ruled out with evidence: pagination (one page, no continuation token), message-count cap (6,715 vs 3,237 vs 2,531), byte cap (134 KB vs 298 KB vs 331 KB), marker formatting, and the CLI itself. The retained window's exact governing unit is still unknown — deliberately, because **nothing here depends on its size.**

## What changed

**Section on `END`, not `BEGIN`.** `END` and the summary print last and always survive. Sections now close on `END` and claim everything since the last boundary, so a head-dropped run still yields real counts. Sections closed without a `BEGIN` are marked partial and warned about.

**One rule, in one place.** `evaluateTestOutcome` is the only definition of a pass; the divergent `_hasJestFailuresInLogs` copy is deleted. It reports *every* reason a run failed, not just the first. Tracebacks always fail in both modes — they catch use-after-free bugs jest cannot see, since a deferred-callback crash fires outside any test.

**Passing requires evidence.** Absence of failure was read as success, so an empty log passed. The check keys on whether jest reported *at all*, not on a test count, so `No tests found` still passes and `passWithNoTests` keeps working. `--script-text` probes are exempt.

**Tracebacks attributed by script path, not log position.** A deferred callback can fire during another package's section — the reason batch mode gave up on tracebacks entirely. Reading the owning package out of the frames fixes that rather than tolerating it. Deduped by signature with counts and bounded, since one bug fires per player per frame.

**Ergonomics:** `--logs` defaults on with `--script-text` (its output is the whole point — the current behaviour was bad enough that a workaround got written into `clienttranslator/docs/engine-behavior.md` instead of a bug report); `batch test --script-text` explains where to go; `--timeout` defaults to the 300 s API maximum.

## Verified against egg-hunt

```
section closed by its END marker with no BEGIN — the start of this output was dropped
Test Suites: 3 failed, 5 skipped, 32 passed, 35 of 40 total
Tests:       19 failed, 27 skipped, 470 passed, 516 total
5 distinct traceback(s)          [400 raw]
✗ FAILED (4m28s)                 EXIT=1
```

Real counts recovered from a head-dropped log, correct red verdict, 400 tracebacks collapsed to 5. **That repository's CI has reported ✅ throughout.**

118 tests pass in `nevermore-cli`; all 5 tools packages build; prettier clean.

## Expect red builds

Any repo with genuinely failing tests goes red on the next run. That is the point, but it is not a quiet change. egg-hunt has 19 real failures, at least 16 of which predate its current branch.

## Not included

#13 per-phase timestamps, #14 threading jest config to the in-engine run, #16 baseline comparison, #17 aligning the documented command with the gate command.

https://claude.ai/code/session_014KaJaDsPyPWQ3whHaav2r8

## Known limitation (found after opening)

A full `batch test --all` sweep cannot pass at large package counts. Measured on Nevermore's 73 test packages: 59 passed, 14 failed, and all 14 failures are `no output could be attributed to this package` for exactly packages 1–14 in run order. The log window cut at `datastore` (package 15), which survived as a partial section and passed 339/339. Every one of those 14 passes with real counts in an 8-package run.

**There are zero real test failures in Nevermore under these rules.** The 14 are unreadable, not broken — and before this PR they would have reported green with no counts behind them, which is worse.

Normal CI is unaffected: `batch test --base origin/main` tests only changed packages, comfortably inside the window. Tracked as a follow-up; the fix is either carrying per-package counts inside `BATCH_TEST_SUMMARY` (which survives by construction) or splitting large batches across multiple execution tasks.

### DIFF
--- a/tools/nevermore-cli/src/commands/batch-command/batch-test-command.ts
+++ b/tools/nevermore-cli/src/commands/batch-command/batch-test-command.ts
@@ -60,64 +60,84 @@ export const batchTestCommand: CommandModule<
   command: 'test',
   describe: 'Run tests for changed packages with test targets',
   builder: (yargs) => {
-    return yargs
-      .option('cloud', {
-        describe: 'Run tests via Open Cloud instead of locally',
-        type: 'boolean',
-        default: false,
-      })
-      .option('api-key', {
-        describe: 'Roblox Open Cloud API key (--cloud only)',
-        type: 'string',
-      })
-      .option('all', {
-        describe: 'Test all packages with test targets, not just changed',
-        type: 'boolean',
-        default: false,
-      })
-      .option('base', {
-        describe: 'Git ref to diff against for change detection',
-        type: 'string',
-        default: 'origin/main',
-      })
-      .option('concurrency', {
-        describe: 'Max parallel tests (0 = unlimited, default: unlimited)',
-        type: 'number',
-      })
-      .option('output', {
-        describe: 'Write JSON results to this file',
-        type: 'string',
-      })
-      .option('limit', {
-        describe: 'Max number of packages to test (for local debugging)',
-        type: 'number',
-      })
-      .option('logs', {
-        describe: 'Show execution logs for all packages (not just failures)',
-        type: 'boolean',
-        default: false,
-      })
-      .option('aggregated', {
-        describe:
-          'Build all packages into a single place and execute one batch script',
-        type: 'boolean',
-        default: true,
-      })
-      .option('batch-place-id', {
-        describe:
-          'Override placeId for the aggregated batch upload (--aggregated only)',
-        type: 'number',
-      })
-      .option('batch-universe-id', {
-        describe:
-          'Override universeId for the aggregated batch upload (--aggregated only)',
-        type: 'number',
-      })
-      .option('timeout', {
-        describe:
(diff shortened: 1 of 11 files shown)


Quenty:
:rocket: PR was released in [`@quenty/nevermore-cli@4.42.0`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli%404.42.0) :rocket:

PR #775, opened 2026-07-29
fix(chatproviderservice): let a PlayerMock be chat tagged

Quenty:
Two mock-safety gaps that stop the chat tag stack booting or being exercised headlessly.

## `PromiseAddChatTag` / `ClearChatTags` asserted a real `Player`

`HasChatTags` is a `PlayerBinder`, which already binds `PlayerMock`s, and every native member the binder touches (`_obj.Name`, parenting the tag container) works on a mock — it is a `Folder` in `Players`. The only thing in the way was the guard. Relaxed to the documented clause:

```lua
assert((typeof(player) == "Instance" and player:IsA("Player")) or PlayerMock.isMock(player), "Bad player")
```

`ChatProviderService.spec.lua` covers it end to end: a mock created through `PlayerMockService` takes a tag through discovery → bind → `AddChatTag`, with the test tagging nothing by hand. The built-in `(dev)`/`(mod)` tags are switched off per test — they resolve through `PermissionService`, which is a different question than whether a mock can be tagged at all.

## `ChatProviderServiceClient.Start` assigned `TextChatService.OnIncomingMessage` unconditionally

The engine rejects that outside a client, so any bag booting the client graph on the server (a headless dual-realm boot test) went down with it. It returns early there now — there is no chat to render in that context anyway.

## Testing

`nevermore test --cloud` in `src/chatproviderservice` — 6/6 (3 pre-existing + 3 new).

Found while adding a demo-player chat tag in a consumer game, which currently carries local workarounds for both of these and drops them once this publishes.

https://claude.ai/code/session_01MJ1WHz5RmWtj7TzZ1jaDrV

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1169,6 +1169,9 @@ importers:
       '@quenty/playerbinder':
         specifier: workspace:*
         version: link:../playerbinder
+      '@quenty/playermock':
+        specifier: workspace:*
+        version: link:../player-mock
       '@quenty/playerutils':
         specifier: workspace:*
         version: link:../playerutils
--- a/src/chatproviderservice/package.json
+++ b/src/chatproviderservice/package.json
@@ -43,6 +43,7 @@
     "@quenty/nevermore-test-runner": "workspace:*",
     "@quenty/permissionprovider": "workspace:*",
     "@quenty/playerbinder": "workspace:*",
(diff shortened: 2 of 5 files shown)


Quenty:
Reopening on `users/quenty/chatprovider-mock-safe` to match this repo's branch naming.

PR #776, opened 2026-07-29
fix(chatproviderservice): let a PlayerMock be chat tagged

Quenty:
Three gaps in the chat tag stack, all found by a consumer game registering `ChatProviderService` for the first time.

## The service never cleaned up after itself

`SetDeveloperTag` / `SetAdminTag` each hold a subscription to **every player in the place**, and `ChatProviderService` had no `Destroy` — so its maid outlived its own `ServiceBag`. A destroyed bag went on resolving permissions and reaching for binders through services it no longer owned.

In a long-lived server that is invisible. In a headless test place it is loud: the leaked subscription keeps firing for PlayerMocks created by later specs, into a Jest module context that has been reset, so the provider's class table is gone and you get

```
attempt to call missing method 'PromiseIsPermissionLevel' of table
```

once per player, forever — which reads as a `PermissionService` bug and is not one.

## `PromiseAddChatTag` / `ClearChatTags` asserted a real `Player`

`HasChatTags` is a `PlayerBinder`, which already binds `PlayerMock`s, and every native member the binder touches (`_obj.Name`, parenting the tag container) works on a mock — it is a `Folder` in `Players`. Only the guard was in the way. Relaxed to the documented clause:

```lua
assert((typeof(player) == "Instance" and player:IsA("Player")) or PlayerMock.isMock(player), "Bad player")
```

## `ChatProviderServiceClient.Start` assigned `TextChatService.OnIncomingMessage` unconditionally

The engine rejects that outside a client, so any bag booting the client graph on the server (a headless dual-realm boot test) went down with it. It returns early there now — there is no chat to render in that context anyway.

## Testing

`nevermore test --cloud` in `src/chatproviderservice` — 9/9 (3 pre-existing + 6 new), covering the tag pipeline end to end: a mock takes a tag through discovery → bind → `AddChatTag` with nothing tagged by hand, the creator takes a `(dev)` tag, nobody else does, and a destroyed bag tags no one.

Worth knowing about that last one: **without the `Destroy` it still passes its assertion and the run reports failed on the traceback** — the same shape the bug took downstream. The cross-spec module-context symptom is not reproducible in-process, so the test pins the lifecycle contract rather than the exotic failure.

_(Supersedes #775, which was opened on a misnamed branch.)_

https://claude.ai/code/session_01MJ1WHz5RmWtj7TzZ1jaDrV

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1169,6 +1169,9 @@ importers:
       '@quenty/playerbinder':
         specifier: workspace:*
         version: link:../playerbinder
+      '@quenty/playermock':
+        specifier: workspace:*
+        version: link:../player-mock
       '@quenty/playerutils':
         specifier: workspace:*
         version: link:../playerutils
--- a/src/chatproviderservice/package.json
+++ b/src/chatproviderservice/package.json
@@ -43,6 +43,7 @@
     "@quenty/nevermore-test-runner": "workspace:*",
     "@quenty/permissionprovider": "workspace:*",
     "@quenty/playerbinder": "workspace:*",
+    "@quenty/playermock": "workspace:*",
     "@quenty/playerutils": "workspace:*",
     "@quenty/preferredparentutils": "workspace:*",
     "@quenty/promise": "workspace:*",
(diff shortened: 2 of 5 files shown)


Quenty:
:rocket: PR was released in [`@quenty/chatproviderservice@9.67.4`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/chatproviderservice%409.67.4) :rocket:

PR #778, opened 2026-07-30
fix(gameproductservice): stop ownership queries spamming the log with failures they cannot avoid

Quenty:
A live server was printing a pair of `[Promise] - Uncaught exception in promise` warnings every few seconds, forever, split between two engine refusals:

```
[Promise] - Uncaught exception in promise: "MarketPlace:UserOwnsGamePassAsync() can only query local player"
[Promise] - Uncaught exception in promise: "Ownership check failed because HTTP 403"
```

Two separate faults, one visible symptom.

## The queries could never succeed

A client holds a `PlayerProductManager` for **every** player — the `PlayerProductManager` tag replicates, so the client binder builds one per player — but only the local player's manager overrides the query callbacks. Everyone else keeps `PlayerProductManagerBase`'s, which ask the marketplace about a player this realm may not ask about: `UserOwnsGamePassAsync` refuses outright, and the inventory endpoints (the paid-access `GAME` lookup among them) answer HTTP 403. One doomed request per player, per asset type.

They are now wired only where they can be answered: any server realm, the local player, or a `PlayerMock` — whose ownership is read through `PlayerMock.readLookup` and never reaches the engine, so headless tests keep driving ownership for players other than the one standing in for the local one. Elsewhere they are left unwired, and an ask rejects naming the missing callback instead of the engine's error.

Nothing that worked stops working: those queries could only ever fail. Ownership overrides are unaffected — they replicate and short-circuit the query, so a client still answers for another player wherever the server has said something — and membership stays wired everywhere, being a replicated property rather than a lookup.

## One failure was reported once per consumer

`PlayerAssetOwnershipTracker` attached two continuations to every query and gave neither a rejection handler: one caching the result, one waiting to observe it. Every `PromiseOwnsAsset` caller and every `ObserveOwnsAsset` subscriber lands on the same cached promise, so **one** failed lookup printed a pair of anonymous warnings **per consumer** — which is what turned two real faults into a wall of log.

Both are handled now:

- the query site reports once, naming the asset type, the id, and what the lookup said
- the observe path consumes silently and emits nothing, so the observable stays silent — unresolved, which is not the same as "does not own it"
- an asset with no callback for this realm is not reported at all: a realm that cannot ask is not a lookup that failed, and its rejection already explains itself

## Tests

`nevermore test --cloud` in `src/gameproductservice`: **87/87 green**.

Seven new cases — five pinning which realm may ask about whom, two pinning that a failed query is reported once however many consumers ask (and that an unresolvable asset emits nothing rather than `false`). Both new assertions were inverted and re-run first to confirm they actually fail, since a green suite here is not by itself evidence.

https://claude.ai/code/session_01AaYt15KuYi84TeHPnhakWc

### DIFF
--- a/src/gameproductservice/src/Shared/Ownership/PlayerAssetOwnershipTracker.lua
+++ b/src/gameproductservice/src/Shared/Ownership/PlayerAssetOwnershipTracker.lua
@@ -51,6 +51,7 @@ export type PlayerAssetOwnershipTracker =
 			_overrideAttribute: any,
 			_appliedOverrideState: any,
 			_assetOwnershipPromiseCache: { [number]: Promise.Promise<boolean> | boolean },
+			_warnedQueryFailure: { [number]: true },
 		},
 		{} :: typeof({ __index = PlayerAssetOwnershipTracker })
 	))
@@ -80,6 +81,7 @@ function PlayerAssetOwnershipTracker.new(
 		JSONAttributeValue.new(self._player, PlayerProductOwnershipOverrideUtils.attributeName(assetType), nil)
 
 	self._assetOwnershipPromiseCache = {}
+	self._warnedQueryFailure = {}
 
 	self._maid:GiveTask(self._marketTracker.Purchased:Connect(function(idOrKey)
 		self:SetOwnership(idOrKey, true)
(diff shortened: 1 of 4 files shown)


Quenty:
Pushed review follow-ups (b838ba4). Three findings were real:

**The tests could not detect the fix being removed.** The fake callback returned an already-rejected promise, and `Then` on a settled promise hands back the parent rather than a child — so no consumer ever held an unhandled rejection, and deleting the `_observeBaseOwnsAssetId` handler kept the suite green. I confirmed that: 87/87 passed with the handler gone. The query is now *pending* when the subscribers attach (the shape the live failure had), and both handlers are counted through named seams — `_warnQueryFailed` and `_consumeObservedQueryFailure`. Deleting either now fails, and it fails by timing out on the consumption rather than by luck.

Worth recording that my first rewrite of that test was itself broken — `expect` received the two values `Yield` returns — so its red proved nothing either. The control was only meaningful once the suite was green (88/88) *and* the deletion turned it red for the right reason.

**`TieRealms.SHARED` opted straight back into the bug.** `~= CLIENT` wires the cloud queries, so any bag told it is shared on a client gets the 403-per-player flood back. The engine's restriction is the *process's*, not the realm's, so the guard now asks the process — mirroring `PlayerProductManagerClient`'s existing `RunService` gate for its bulk-purchase listener — behind an `_isClientProcess` seam a spec can drive, since `--cloud` runs server-side. That removes the realm from the question entirely rather than adding a SHARED branch to it.

**`_warnedQueryFailure` was dead state.** Its only writer cleared it in the same breath as the query cache, so the guard was never true when read; the comment justifying it was describing the cache's job. Dropped, and the method renamed to `_warnQueryFailed`.

Also taken: `--[=[ ]=]` with `@param`/`@private` on the new private methods to match the file's neighbours, the class patch restored in `afterEach` (a test killed mid-yield would otherwise leave the recorder installed for the run), and the two inaccurate comments deleted — including the claim that the check order was load-bearing, which it isn't, the predicate being a pure disjunction.

**Not taken, deliberately:** evicting the cache entry on rejection so a transient marketplace failure can be retried. That behaviour is unchanged by this PR (a rejected query has always been cached for the tracker's lifetime, which is why `OwnsGameAccessFact` can abstain permanently), and adding retries without backoff would re-open the request volume this PR set out to reduce. It deserves its own change.

88/88 green in `src/gameproductservice`; stylua, selene and luau-lsp clean.

Quenty:
:rocket: PR was released in [`@quenty/access@1.3.7`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/access%401.3.7), [`@quenty/gameproductservice@14.66.5`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/gameproductservice%4014.66.5) :rocket:

PR #780, opened 2026-07-30
fix(datastore): stop manager teardown from destroying stores before PlayerRemoving saves them

Quenty:
Player data was being lost on server restarts because destroying the manager tore each datastore down before PlayerRemoving could save it, and Roblox fires PlayerRemoving for everyone still in a closing server. Depending on which version of that teardown flush ran, a restart either left the session lock held (costing the next server about 76 seconds before it could steal it) or destroyed the store while players were still writing to it.

Player saves now go through the one path they went through before, and the close path waits for any removal still in flight rather than a snapshot that can be empty while a save is mid-write.

The specs had been modelling a shutdown as destroying the manager, which is not what a shutdown is, so they now simulate PlayerRemoving followed by the close the way a real server does.

## Also: the crash that lock cost surfaced (saveslot)

A production server hit `HasSaveSlots:1113: attempt to call missing method '_buildSlot' of table` while retrying a load against exactly the held lock described above. `BaseObject.Destroy` strips the metatable, so that message is a method call on a destroyed object: the player left during the retry ladder, the binder was destroyed, and when the lock finally cleared the load's continuation called into it. Because a load resolves inside its own `UpdateAsync` transform, Roblox reported it as a `Transform function error` and discarded that transform's write.

`HasSaveSlots._promiseLoadSlots` was the one load path f8716a37 missed when it fixed this same defect in `SaveSlotService`'s selection chain, so it now takes the same shape: every hop maid-owned, every continuation re-checking the object is alive. `SaveSlotLateSettle.spec` had only covered a fresh player, whose load settles with no metadata and never reaches `_buildSlot`; it now seeds a returning player's slots, which reproduces the production stack at the same line without the guards.

Verified on `src/saveslot` via Open Cloud: 189/189 with the guards, and the same suite exits 1 on that stray error without them.

https://claude.ai/code/session_01NLMCUpaUds4ZuB9o4bkgXX
https://claude.ai/code/session_01HouVsNhp1j387CUHa4zCMq

### DIFF
--- a/src/datastore/README.md
+++ b/src/datastore/README.md
@@ -18,6 +18,16 @@ This system is a reliable datastore system designed with promises and asyncronio
 ## Executive overiew
 This datastore prevents data loss by being explicit about what we're writing to, and only modifying the data that exists there instead of modifying the whole structure.
 
+## Working on this package
+
+Read this before changing how player data is saved, and before adding any cleanup, teardown, or
+flush to that path. It records design intent that cannot live in the code, because the shape is
+driven by Roblox shutdown and cross-server session-lock behavior that is not visible from reading it:
+
+- [`docs/shutdown-and-session-locks.md`](docs/shutdown-and-session-locks.md) — why every save routes
+  through one function, what `PromiseAllSaves` has to wait on, and why flushing stores on manager
+  teardown was tried twice and removed.
+
 ## Comparison to other solutions
 
 * Not specifically locked to players
--- a/src/datastore/docs/shutdown-and-session-locks.md
+++ b/src/datastore/docs/shutdown-and-session-locks.md
@@ -17,11 +17,16 @@ shutdown, not a peacetime-only path that something else has to stand in for.
 
 **A session lock can only be released by the server holding it.** Nothing gives one server the
 authority to unlock another's key. If a server dies still holding a lock, the next server that loads
-that key sees a lock that still looks live, and takes the graceful route: it messages the holder's
-JobId asking it to close. That JobId is gone, so nothing answers, and it waits out
-`SetSessionMessagingCloseDelaySeconds` (5s default) before retrying — six attempts, ~76 seconds,
-against a holder that no longer exists — and only then steals the lock. The player waits through all
-of it.
+that key sees a lock that still looks live and takes the graceful route: it messages the holder's
+JobId asking it to close, and that JobId is gone, so nothing answers.
+
+The cost of that is worth knowing precisely, because the obvious knob is the wrong one. Each attempt
+dies on the **hardcoded 5s timeout in `DataStoreMessageHelper.PromiseCloseSessionGraceful`**, not on
+`SetSessionMessagingCloseDelaySeconds` — that delay sits in the *fulfilled* branch of
+`_promiseGetAsyncNoCache` and is never reached when the holder is dead. Six attempts of that is 30s,
+and `PromiseRetryUtils` adds five jittered inter-attempt waits totalling ~49s, so a player waits
+roughly **80 seconds** before the lock is finally stolen. Turning `SetSessionMessagingCloseDelaySeconds`
+down does not shorten it; the lever is that hardcoded 5.
 
 ## The shape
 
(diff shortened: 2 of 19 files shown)


Quenty:
### Subagent review of the saveslot half — outcome

Verified sound and left as-is: the `if not self.Destroy` liveness test (no `Destroy` override anywhere in the `HasSaveSlots` → `HasSaveSlotsBase` → `BaseObject` chain, never an own field, so it resolves only through the metatable); and the `return nil` branches, which cannot fulfil `_loadPromise` with nil — `DoCleaning` rejects the line-107 wrapper *before* `setmetatable(self, nil)`, so the later resolve hits a promise whose `_pendingExecuteList` is already nil and consumers still see a value-less rejection. Rejecting with no values is non-reportable, so this adds no `[Promise]` warning noise.

Acted on in 2b76762:

- **`PromiseSelectTransferableEphemeralSlot` had the same defect** — raw shared-store read, continuation calls the same `_buildSlot`. It is on the join path (`SaveSlotService` starts it on every bind), so it is now maid-owned too. One correction to the review's reasoning: it argued the read is triggered *by* the leave via `TeleportDataService._getEntry`'s cleanup, but that cleanup resolves with `nil`, so a player leaving pre-replication bails at the `type(key) ~= "string"` check. The real trigger is the ordinary one — leaving during the `GetAsync` of a player who did arrive with a key.
- **The regression test could have rotted silently.** It asserted only that the load was still pending, so drift in the seeded raw layout would have reduced it to the empty-metadata case it exists to replace — green, and no longer reaching `_buildSlot`. Added a stay-put test that reads the seed back through `PromiseHasSlot` / `PromiseLastActiveSlotId`.
- Comment now states that maid ownership is the mechanism and the liveness checks are belt-and-braces.

Knowingly **not** in scope here, and worth a follow-up rather than widening a shutdown PR: the same raw-hop shape survives in `PromiseDuplicateSlot`, `PromiseResetSlot`, `PromiseImportSlotFromSharedDataStore`, `HasSaveSlots.lua:~225` (`_dataStore:Save()` → pre-select), and `HasSaveSlotsBase.PromiseLoadSaveSlotFromTeleport`. Those are player-initiated mid-session actions rather than the join/leave window this PR is about. The right fix is one `HasSaveSlotsBase` helper (maid-own + liveness in a single place) plus a guard test, not a seventh hand-copy of the idiom.

`src/saveslot` via Open Cloud: **190/190, exit 0, no stray errors.** Reverting the guards puts the production stack back at the same line and exits 1.

Quenty:
### Second review round — SHIP verdict, four fixes applied (14d551b)

A re-review of the saveslot half returned **ship**, with one gap I had wrongly waved off as out of scope:

- **`PromiseImportSlotFromSharedDataStore` had the same defect one function above the one I'd just fixed** — and is *more* exposed, because unlike the ephemeral path it isn't gated behind `_loadPromise`, so nothing upstream rejects it when the player leaves mid-read. Now maid-owned + liveness-checked. My earlier "player-initiated mid-session, out of scope" call was wrong on the facts.
- **Both shared-store paths are declared `Promise<SlotId>`, so the destroyed branch now rejects instead of fulfilling nil** — a nil fulfilment read as "imported, no slot" to `SaveSlotCmdrService`, which printed a false success. The rejection is deliberately **empty**: `Promise` only reports rejections carrying information, and neither caller has a `Catch`, so a message would surface as an uncaught-exception traceback and fail the `--cloud` gate.
- **The stay-put test now waits on `ActiveSlotId` itself** rather than `PromiseLastActiveSlotId`, which can answer from `_lastActiveSlotId` while the selection is still in flight — pins the selection directly and quiesces the chain before teardown.
- **Comment corrected.** Continuations queued on a maid-owned promise before the maid dies receive its *rejection*, so the liveness checks don't cover "a continuation queued before the maid died" as I'd written; what they cover is a settle landing mid-teardown, before that promise's own maid task came up.

Independently verified by the reviewer, worth recording: the positive control is genuine — `_slotMap` is populated *only* by `_buildSlot` and `EXISTING_SLOT_ID` is unreachable from any creation path, so `PromiseHasSlot` returning true proves the seed was read *and* `_buildSlot` ran on it. `_buildSlot`'s `self._maid[slotId] = maid` is also safe mid-cleaning, since `Maid.DoCleaning` drains tasks added during the drain. No later-suite isolation hazard from the added test.

Still out of scope, deliberately: `PromiseDuplicateSlot`, `PromiseResetSlot`, `HasSaveSlots.lua:~225`, and `HasSaveSlotsBase.PromiseLoadSaveSlotFromTeleport` carry the same shape behind mid-session actions. Those want the shared `HasSaveSlotsBase` helper plus a guard test, not a hand-copy each.

`src/saveslot` via Open Cloud after all four: **190/190, exit 0, no stray errors or uncaught-exception warnings.**

Quenty:
:rocket: PR was released in [`@quenty/chatproviderservice@9.67.5`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/chatproviderservice%409.67.5), [`@quenty/datastore@13.51.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/datastore%4013.51.1), [`@quenty/saveslot@2.6.2`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/saveslot%402.6.2), [`@quenty/secrets@7.64.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/secrets%407.64.1), [`@quenty/settings-inputkeymap@10.70.4`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/settings-inputkeymap%4010.70.4), [`@quenty/settings@11.67.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/settings%4011.67.1), [`@quenty/softshutdown@10.3.5`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/softshutdown%4010.3.5) :rocket:

PR #785, opened 2026-08-04
fix(saveslot): carry accrued playtime onto a copied slot

buildthomas:
## Problem

Copying a save slot loses its playtime — the copy shows as 0 time played even though it carries the same progress.

Both copy paths seed the new slot with only `SlotName` and `Summary`:

- `HasSaveSlots.PromiseDuplicateSlot` passes those two to `PromiseCreateSlot`
- `PromiseExportSlot` builds a `SaveSlotExport` of `{data, slotName, summary}`, and `PromiseImportSlot` seeds the new slot from those

`TimePlayed` is never part of that hand-off, so it starts nil on every copy.

## Fix

Playtime travels with the progress it measures:

- `SaveSlotCreateMetadata` takes an optional `TimePlayed`, applied in `PromiseCreateSlot` (it is already one of `MUTABLE_METADATA_KEYS`, so `_buildSlot` persists it like any other metadata field)
- `PromiseDuplicateSlot` and `PromiseImportSlot` pass the source's value through
- `SaveSlotExport` gains `timePlayed`, written by `PromiseExportSlot` and validated as an optional number by `isSaveSlotExport`
- Both duplicate and export call `_flushPlaytime()` first, so copying the *active* slot includes the running session rather than only what the last save wrote

Timestamps and session counters (`PlayCount`, `LastSessionLength`) still start fresh on a copy — those describe the new slot's own sessions, not the progress being copied.

## Backwards compatibility

- Exports written before this change have no `timePlayed`; they still validate and import to a slot with no accrued playtime, exactly as they did when written
- A newer export read by older code just carries an unused field
- No change to how slot metadata is stored or loaded: `TimePlayed` was already an optional attribute, this only seeds it at creation
- `SaveSlotExportUtils.create` gained a trailing optional parameter, so existing 3-argument callers are unaffected

`TimePlayed` in the create metadata is not a new trust surface: `PromiseSetSlotMetadata` is already client-callable and accepts the full `SaveSlotMetadata`, `TimePlayed` included.

## Tests

New coverage in `HasSaveSlots.spec.lua` (duplicate carries playtime including the unflushed live session; a never-played slot's copy still has none), `HasSaveSlots.ExportImport.spec.lua` (export/import round-trips playtime, exports the live session, imports a pre-change export, refuses a non-numeric `timePlayed`), and `SaveSlotExportUtils.spec.lua` (create/validation).

stylua and selene pass. I could not run `nevermore test --cloud` here — no Open Cloud key in this environment, and the local `rojo` shim is broken on this machine — so the new specs have not been executed against the engine.


alex-y-z:
Merged in https://github.com/Quenty/NevermoreEngine/pull/793

PR #790, opened 2026-08-08
fix(gameproductservice): verify gamepass ownership before granting a client-reported purchase

buildthomas:
The server trusted the client-fired PromptGamePassPurchaseFinished remote, so firing it with isPurchased = true for any gamepass id granted ownership without the player ever paying (the standing "TODO: Validate in purchased scenario"). The pass purchase is now confirmed server-side with MarketplaceService.UserOwnsGamePassAsync before it is recorded, so an unverified or unconfirmed claim grants nothing, and a marketplace error grants nothing rather than trusting the client. Adds dual-realm spec coverage for both the confirmed-purchase and unconfirmed-claim paths.


alex-y-z:
Fixed in https://github.com/Quenty/NevermoreEngine/pull/791

PR #794, opened 2026-08-10
fix(nevermore-cli): stop failing a batch run whose tests all passed

Quenty:
A cloud batch run whose tests all passed could be reported as FAILED with "no output could be attributed to this package". The log scan stopped at the summary marker, so a section whose END arrived after the summary was thrown away, and the sort that put the END there compared timestamps as strings even though the API emits a variable number of fractional digits. The scan now continues past the summary, and the sort is removed rather than fixed, since probes show messages already arrive in order and a whole frame of prints shares one timestamp anyway.


Quenty:
:rocket: PR was released in [`@quenty/nevermore-cli@4.43.1`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli%404.43.1) :rocket:

PR #795, opened 2026-08-11
Return a test run's results instead of scraping them out of its logs

Quenty:
A test run now returns a table with its verdict, test and suite counts, and a capped list of failed test names, instead of announcing a failing suite by throwing and leaving the counts to be scraped back out of log text that Open Cloud truncates on long runs. A script's return value travels to the CLI on ScriptRunResult.returnValues, filled in by the Open Cloud transport from the task's output.results and by the local transport from what the Studio bridge plugin marshals off the executed script. The field is absent rather than empty when a transport delivered no result at all, so a reader can tell "nothing came back" apart from "the script returned nothing" and fall back to log parsing only in the first case. This also corrects LuauTask.output, which claimed a { value?: string } wrapper that Open Cloud never sends.

No test scripts or scaffolding templates change here, so every package still returns nothing and every run still takes the log-scraping path it took before. Both readers require a jest report when no results came back, so a run they cannot read fails rather than passing quietly. Packages opt into returning their results one at a time, and get their counts back exactly when they do.

Because the plugin now walks values the script chose, the execute action marshals them defensively (raw traversal, cycle and depth caps, a node budget, markers for non-finite numbers and colliding keys) and each queued execute request runs under pcall — a throw in that window used to latch the queue for the rest of the Studio session, which no later request could recover from.

This supersedes #797, which GitHub would not let retarget main because it was locked as part of a stack.

Worth a separate look, same failure mode and out of scope here: query-data-model.luau's serializeValue has no nan/inf guard, so a NaN attribute encodes into a message the server cannot decode and the response is lost.

PR #796, opened 2026-08-11
refactor(nevermore-cli): split batch log parsing into tokenize then fold

Quenty:
Batch log parsing now classifies each log line into a token and folds the token list into per-package sections, instead of threading twelve mutable bindings through a single loop where the diagnostic counters also decided the parse. Per-package results are byte-for-byte the same; the first commit adds characterization tests that pin the previously untested behavior, and the second commit is the restructure. Two anomalies that used to pass silently are now warned about without changing any verdict: a second BEGIN arriving while a section is still open, and an END naming a slug that is not in this batch.

### DIFF
--- a/tools/nevermore-cli/src/utils/testing/parsers/batch-log-parser.test.ts
+++ b/tools/nevermore-cli/src/utils/testing/parsers/batch-log-parser.test.ts
@@ -151,6 +151,241 @@ describe('parseBatchTestLogs', () => {
     expect(result?.testCounts?.failed).toBe(0);
     expect(result?.tracebackCount).toBe(1);
   });
+
+  it('fails a section containing a traceback', () => {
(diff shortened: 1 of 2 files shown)


Quenty:
Cherry-picked into #795 (commits 60e95e6ed2 and 40b0ec9480), where the tokenize step is the substrate for rendering batch output line by line — every line becomes a token, so output between packages survives instead of being dropped at the old parse loop.

Two of the characterization tests here pinned wording that #795 deliberately changed: a failing suite now reaches the batch summary as a verdict carrying the runner's own reason, so 'the batch runner reported a Luau error' became that reason ('boom' in the one case, 'the batch runner reported this as failed' in the other). Everything else came over unchanged and all 42 tests pass alongside #795's.

PR #797, opened 2026-08-11
Return a test run's results instead of scraping them out of its logs

Quenty:
A test run now returns a table with its verdict, test and suite counts, and a capped list of failed test names, instead of announcing a failing suite by throwing and leaving the counts to be scraped back out of log text that Open Cloud truncates on long runs. `nevermore test` and `nevermore batch test` read that verdict, so a failing suite is detected structurally rather than by finding a jest summary line that may not have survived truncation. Every package's test script was updated to return its results; one that has not been updated returns nothing and behaves exactly as before.


Quenty:
Collapsed into #795, which now carries the whole change against main. GitHub would not let this PR retarget main (locked as part of a stack), so the work moved onto the other branch and this one closed itself out.

PR #806, opened 2026-08-16
fix(nevermore-cli): install with the project's package manager

unrooot:
`nevermore install` always ran `npm install`, which fails outright in a pnpm project because several published @quenty packages still carry a `workspace:*` range in their devDependencies and npm rejects that protocol. It now installs with whatever package manager the project uses, detected from a `packageManager` field or a lockfile, defaulting to pnpm when a directory declares neither, and passes `add` rather than `install` to everything but npm so the dependency is actually recorded.

Package name validation was also rejecting valid packages: it checked names against a single npm registry search request, which caps at 250 results while 300+ @quenty packages are published, so anything in the tail (metricutils, camerainfo, safedestroy, and roughly fifty others) failed as "Invalid packages". Names are now checked against the registry directly, and an unreachable registry no longer looks like a typo.

Verified end to end against a real pnpm project and a bare directory; `nevermore install blend metricutils` now succeeds where both halves previously failed.


Quenty:
:rocket: PR was released in [`@quenty/nevermore-cli@4.43.4`](https://github.com/Quenty/NevermoreEngine/releases/tag/%40quenty/nevermore-cli%404.43.4) :rocket:

